PR #512, opened 2024-09-20
Switch from dev to rc

saeliddp:
## Context
After some research, @hughes036 and I found that the `rc` tags are most commonly used to indicate pre-releases (not `dev`, which we've been using).

## Changes
1. update bumpver config in `pyproject.toml`
2. update `bumpver_handler.py` to use rc instead of dev
3. update options in bumpversion GH workflow

## Testing
Updated tests for `bumpver_handler.py`

### DIFF
--- a/src/allencell_ml_segmenter/scripts/bumpver_handler.py
+++ b/src/allencell_ml_segmenter/scripts/bumpver_handler.py
@@ -11,7 +11,7 @@ def main() -> None:
         raise ValueError("No component specified for bumping version")
 
     component: str = sys.argv[1].lower()
-    valid_options: set[str] = {"major", "minor", "patch", "dev", "post"}
+    valid_options: set[str] = {"major", "minor", "patch", "rc", "post"}
 
     if component not in valid_options:
         raise ValueError(f"Component must be one of {valid_options}")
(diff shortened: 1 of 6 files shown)


src/allencell_ml_segmenter/scripts/bumpver_handler.py
```diff
@@ -20,36 +20,33 @@ def main() -> None:
     version_components: list[str] = version.split(".")
 
     update_output: subprocess.CompletedProcess
-    # 4 components means we currently have a dev or post version
-    if len(version_components) == 4 and version_components[-1].startswith(
-        "dev"
-    ):
-        if component == "dev":
-            # increment the dev tag (e.g. 1.0.0.dev0 -> 1.0.0.dev1)
+    if "rc" in version_components[-1]:
```
hughes036, line 23: Why don't we need to check the # of components now?

hughes036: Seems like that should happen right after we split the string?

src/allencell_ml_segmenter/scripts/bumpver_handler.py
```diff
@@ -60,11 +57,6 @@ def main() -> None:
                 ["bumpver", "update", f"--{component}", "-n"]
             )
 
-    else:
-        raise ValueError(
```
hughes036, line 64: Also wondering about this change - can you explain the refactor?

src/allencell_ml_segmenter/scripts/bumpver_handler.py
saeliddp: The standard for RC seems to be `X.X.XrcY` (note the lack of a [period between patch and RC). Bumpver, for some reason, makes you choose a pattern for all tags to match, so in order to get `rc` in the right place, we have to make both rc and post tags follow the no-period pattern. This means that the number of components derived by splitting on periods will always be three now. 

I could also do a fancier regex to split on `(\.|[a-z])+`, but then I'd have to go back to look at the original string anyways to figure out whether its rc or post.

yrkim98 (approved):

amilworks (approved):

hughes036 (approved):

PR #513, opened 2024-09-23
absolute url to license, BSD -> AISL in classifiers

hughes036:
- The metadata on our pypi index page was still displaying BSD as the license, so I updated it to display AISL.
- The link to our license file on the readme was relative, so I updated to to be absolute.  Hopefully this will make it resolvable from PYPI.

The last thing I need to figure out is where this section on the index is rendered from:
image
Any ideas @amilworks ?  I already changed every reference to BSD in the repo.


amilworks (approved):
Link works, LGTM.

amilworks:
Is this what you are asking @hughes036 

https://github.com/AllenCell/allencell-segmenter-ml/blob/b5e5f072189dd56a1c4f8ec75c425c1410487b28/pyproject.toml#L26

hughes036:
> Link works, LGTM.
> 
> Isn't this rendered from the `pyproject.yaml`?

Maybe, but if so I wonder how how it generates this value for `License` That value doesnt exist in pyproject.yaml.  Maybe it parses the LICENSE files?

amilworks:
> Maybe, but if so I wonder how how it generates this value for `License` That value doesnt exist in pyproject.yaml. Maybe it parses the LICENSE files?

Okay, after looking around other GitHub repos, I agree with your hunch that it probably parses the BSD License from the `LICENSE` file. Either that or just change the `pyproject.yaml` and see what happens.

kmitcham (approved):

pyproject.toml
```diff
@@ -23,7 +23,7 @@ classifiers = [
     "Development Status :: 2 - Pre-Alpha",
     "Framework :: napari",
     "Intended Audience :: Developers",
-    "License :: OSI Approved :: BSD License",
```
pgarrison, line 26: Trove classifiers must come from [the shared list](https://pypi.org/classifiers/). For the AISL, the following seem appropriate:
```
License :: Other/Proprietary License
License :: Free for non-commercial use
```

pgarrison (changes requested):

pyproject.toml
hughes036: ca56c801123e80e8574dbf6d4e7b942b4bb91d91

thao-do (approved):

tyler-foster (approved):

tyler-foster (approved):

PR #517, opened 2024-09-24
publishing from public GHA runners

hughes036:
Until now our publish workflow ran on our self-hosted runner so that it could PUT artifacts in Artifactory (private PYPI).  That is disallowed from public repos.  Now that this repo is public, we must run workflows on the public runners.  This means we can only publish to PYPI.


amilworks (approved):
Nice, self-hosted runner will speed up our other workflows.

PR #519, opened 2024-09-27
Bugfix/enforce model trained once

yrkim98:
UI changes to enforce that the user can only train a model once. Will automatically switch to prediction tab and disable curation/training once plugin is able to verify a model was sucessfully trained.

Changes:

- src/allencell_ml_segmenter/main/main_model.py: Dispatch training complete event
- src/allencell_ml_segmenter/main/main_widget.py: Subscribe to event and disable tabs when training complete
- src/allencell_ml_segmenter/training/view.py: Add check to see if training completed successfully, and dispatch event if so

### DIFF
--- a/src/allencell_ml_segmenter/main/main_model.py
+++ b/src/allencell_ml_segmenter/main/main_model.py
@@ -85,4 +85,3 @@ class MainModel(Publisher):
 
     def training_complete(self) -> None:
         self.dispatch(Event.PROCESS_TRAINING_COMPLETE)
-
--- a/src/allencell_ml_segmenter/main/main_model.py
+++ b/src/allencell_ml_segmenter/main/main_model.py
@@ -82,3 +82,7 @@ class MainModel(Publisher):
         ):
             self._selected_channels = new_channels
             self.signals.selected_channels_changed.emit()
+
+    def training_complete(self) -> None:
+        self.dispatch(Event.PROCESS_TRAINING_COMPLETE)
+
--- a/src/allencell_ml_segmenter/main/main_widget.py
+++ b/src/allencell_ml_segmenter/main/main_widget.py
@@ -148,7 +148,9 @@ class MainWidget(AicsWidget):
 
         # events for auto window switching
         self._model.subscribe(
-            Event.PROCESS_TRAINING_COMPLETE, self, self._disable_non_prediction_tabs
+            Event.PROCESS_TRAINING_COMPLETE,
+            self,
+            self._disable_non_prediction_tabs,
         )
 
     def _handle_experiment_applied(self, _: Event) -> None:
--- a/src/allencell_ml_segmenter/main/main_widget.py
+++ b/src/allencell_ml_segmenter/main/main_widget.py
@@ -148,7 +148,7 @@ class MainWidget(AicsWidget):
 
         # events for auto window switching
         self._model.subscribe(
-            Event.PROCESS_TRAINING_COMPLETE, self, self._handle_existing_model
+            Event.PROCESS_TRAINING_COMPLETE, self, self._disable_non_prediction_tabs
         )
 
     def _handle_experiment_applied(self, _: Event) -> None:
@@ -174,7 +174,7 @@ class MainWidget(AicsWidget):
             else self._prediction_view
         )
 
-    def _handle_existing_model(self, _: Event) -> None:
+    def _disable_non_prediction_tabs(self, _: Event) -> None:
         """
         Handle existing model selection (disable tabs).
 
--- a/src/allencell_ml_segmenter/main/main_widget.py
+++ b/src/allencell_ml_segmenter/main/main_widget.py
@@ -183,8 +183,8 @@ class MainWidget(AicsWidget):
         inputs:
             is_new_model - bool
         """
-        self._window_container.setTabDisabled(0, self._model.is_new_model())
-        self._window_container.setTabDisabled(1, self._model.is_new_model())
+        self._window_container.setTabEnabled(0, False)
+        self._window_container.setTabEnabled(1, False)
 
     def _handle_change_view(self, event: Event) -> None:
         """
--- a/src/allencell_ml_segmenter/main/main_widget.py
+++ b/src/allencell_ml_segmenter/main/main_widget.py
@@ -146,6 +146,13 @@ class MainWidget(AicsWidget):
         layout.addStretch(100)
         self.setStyleSheet(Style.get_stylesheet("core.qss"))
 
+        # events for auto window switching
+        self._model.subscribe(
+            Event.PROCESS_TRAINING_COMPLETE,
+            self,
+            self._handle_existing_model
+        )
+
     def _handle_experiment_applied(self, _: Event) -> None:
         """
         Handle the experiment applied event.
@@ -168,6 +175,16 @@ class MainWidget(AicsWidget):
             if self._model.is_new_model()
             else self._prediction_view
         )
+
+    def _handle_existing_model(self, _: Event) -> None:
+        """
+        Handle existing model selection (disable tabs).
+
+        inputs:
+            is_new_model - bool
+        """
+        self._window_container.setTabDisabled(0, self._model.is_new_model())
+        self._window_container.setTabDisabled(1, self._model.is_new_model())
 
     def _handle_change_view(self, event: Event) -> None:
         """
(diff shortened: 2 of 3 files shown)


src/allencell_ml_segmenter/main/main_widget.py
```diff
@@ -169,6 +174,16 @@ def _handle_new_model(self, _: Event) -> None:
...
         )
 
+    def _handle_existing_model(self, _: Event) -> None:
+        """
+        Handle existing model selection (disable tabs).
+
+        inputs:
+            is_new_model - bool
+        """
+        self._window_container.setTabEnabled(0, False)
```
hughes036, line 184: Dont we have a mechanism already switching between tabs?

src/allencell_ml_segmenter/main/main_model.py
```diff
@@ -82,3 +82,6 @@ def set_selected_channels(
         ):
             self._selected_channels = new_channels
             self.signals.selected_channels_changed.emit()
+
+    def training_complete(self) -> None:
+        self.dispatch(Event.PROCESS_TRAINING_COMPLETE)
```
hughes036, line 87: Is there not already an event being dispatched when training is complete?  I would think that there is because of the progress dialog behavior?

src/allencell_ml_segmenter/main/main_widget.py
hughes036: I am thinking of this https://github.com/AllenCell/allencell-segmenter-ml/blob/main/src/allencell_ml_segmenter/main/main_model.py#L48

src/allencell_ml_segmenter/training/view.py
```diff
@@ -296,18 +297,46 @@ def getTypeOfWork(self) -> str:
         return "Training"
 
     def showResults(self) -> None:
```
hughes036, line 299: I see no references to `showResults`.  How does it get called?

src/allencell_ml_segmenter/training/view.py
```diff
@@ -296,18 +297,46 @@ def getTypeOfWork(self) -> str:
...
-        min_loss: Optional[float] = FileUtils.get_min_loss_from_csv(csv_path)
-        if min_loss is None:
-            raise RuntimeError("Cannot compute min loss")
-        dialog_box = InfoDialogBox(
-            "Training finished -- Final loss: {:.3f}".format(min_loss)
+        if exp_path is None:
+            raise ValueError(
+                "Experiments path should not be None after training complete."
+            )
+        exp_name: Optional[str] = self._experiments_model.get_experiment_name()
```
hughes036, line 308: Are these new validation check related to the pr's focus?  Or just something you noticed we should have?

src/allencell_ml_segmenter/training/view.py
```diff
@@ -296,18 +297,46 @@ def getTypeOfWork(self) -> str:
...
+                csv_path
+            )
+            if min_loss is None:
+                raise RuntimeError("Cannot compute min loss")
+
+            dialog_box = InfoDialogBox(
+                "Training finished -- Final loss: {:.3f}".format(min_loss)
+            )
+            dialog_box.exec()
+            self._main_model.training_complete()
```
hughes036, line 334: I wonder about tying this event to the show results dialog, rather than just rely on a single TRAINING_COMPLETE event for both showing the dialog and changing the tab.

src/allencell_ml_segmenter/main/main_widget.py
yrkim98: I use this mechanism to switch tabs already. this is just to disable the tabs after training completes. I'll rename for clarity.

src/allencell_ml_segmenter/training/view.py
yrkim98: see `/core/view.py`: https://github.com/AllenCell/allencell-segmenter-ml/blob/9756bd3565fedf27702cedfc1af4151da196e6d3/src/allencell_ml_segmenter/core/view.py#L90-L94

src/allencell_ml_segmenter/training/view.py
yrkim98: This is to satisfy mypy requirements

src/allencell_ml_segmenter/main/main_widget.py
yrkim98: https://github.com/AllenCell/allencell-segmenter-ml/blob/852b8c8e6894bc71f9e85c53e0cce7f88eef6998/src/allencell_ml_segmenter/prediction/view.py#L109-L113

hughes036: Shouldnt disabling / enabling be part of the same mechanism as selecting a tab?  (is the behaviors of selecting bound with disabling the other tabs?)

src/allencell_ml_segmenter/training/view.py
yrkim98: So use this even to show the result dialog? Instead of showing it in the training UI code and emitting the event afterwards?

src/allencell_ml_segmenter/training/view.py
hughes036: punting to the refactor ticket (discussed in person)

src/allencell_ml_segmenter/training/view.py
hughes036: Ok.  Maybe we should rename showResults.
Also, for now just comment what `training_complete` is about

hughes036 (approved):

src/allencell_ml_segmenter/training/view.py
yrkim98: https://github.com/AllenCell/allencell-segmenter-ml/pull/519/commits/d2f67edf42fa5ca9e3d8cff5fba62a3b62e5003b

hughes036: https://github.com/AllenCell/allencell-segmenter-ml/issues/521

src/allencell_ml_segmenter/training/view.py
hughes036: https://github.com/AllenCell/allencell-segmenter-ml/issues/522

src/allencell_ml_segmenter/main/main_widget.py
hughes036: https://github.com/AllenCell/allencell-segmenter-ml/issues/523

amilworks (approved):
Just so I understand correctly, this is to ensure that after someone has trained a model through the plugin, they do not mistakenly retrain again. So this disables the tab for training and puts them directly to the prediction tab?

How does this work for iterative training? Say I trained my model, I run prediction, results look good. Now I want to retrain again on new data, but load the model I just trained and continue training from there, do I need to close the Segmenter ml plugin since training is disabled? Or is it easy to just start a new run? Sorry for the weird phrasing, I hope this makes sense.

hughes036:
> How does this work for iterative training? Say I trained my model, I run prediction, results look good. Now I want to retrain again on new data, but load the model I just trained and continue training from there, do I need to close the Segmenter ml plugin since training is disabled? Or is it easy to just start a new run? Sorry for the weird

Good questions.  I think it is to be in line with the existing workflow for newly trained models (which can't be iteratively trained, per the UX design).  Allowing iterative training on the same model was going to open a can of worms in terms of populating the UI correctly and validating what is entered to be coherent with the earlier rounds of training (I think this was the reason - can you confirm @yrkim98 @thao-do )

The workflow that we ended up with (initially during a conversation with Susan) was that for every iteration of training, the user will create a new model based on the weights that they are 'iterating' on.  So in your scenario, rather than doing additional training on `model 1` after a prediction run, they create `model 2` (a new model) using the weights from `model 1` and do a single round of training on that.

Did I get it right @yrkim98 ?

hughes036:
@yrkim98 @amilworks How about we discuss this @ standup?

thao-do:
@hughes036 I think we'll still have to deal with that can of worm to a certain extent - we need to check with @benjijamorris if any UI input will be in conflict with the model weight being used (mainly the Patch size). 

Also, the Training tab was currently disabled in the existing model route since we couldn't quite get iterative training earlier on (I need to look back on the reasons why to see whether they're still a concern). This new addition to pull model weight is a first step toward allowing more than one case of iterative training. We'll need more discussion on this as there are a few things we might want to reassess.

PR #520, opened 2024-09-27
Update issue templates

yrkim98:
Creating a feature request issue template for taking features from users/non dev team members

It asks them 3 questions related to their feature request, and will automatically apply the `requests` label and the `needs prioritization` label when the issue is submittted.

I can create a separate project board for these requests, and we can track them separate from our dev tickets.

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -1,0 +1,23 @@
+---
+name: Feature request
+about: Suggest an idea for this project
+title: ''
+labels: needs prioritization, Requests
+assignees: ''
+
+---
+
+**Please describe the problem that needs to be solved by this feature**
+A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,23 @@
+---
+name: Feature request
```
hughes036, line 2: Never seen this before - will it be used for every new issue on the repo?

yrkim98: You'll have the option to choose. I want to make it so that devs are the only ones that can open regular tickets, but not sure if thats possible. We can just inform users and the non-dev team to open tickets as feature requests, so they're visible to us on a separate board which we can occasionally prioritize/decide if we want to do that work

hughes036 (approved):

saeliddp (approved):

saeliddp:
Should we also have a default template?

PR #524, opened 2024-10-07
Adding version pin and staging bucket

yrkim98:
Need new models + updated cyto-dl for iterative training. This will fix things for thao's tests.

Just updated version pin and set STG_BUCKET as bucket for models.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,7 +41,7 @@ dependencies = [
     "bioio",
     "tifffile>=2023.4.12",
     "watchdog",
-    "cyto-dl>=0.1.8",
+    "cyto-dl>=0.3.0",
     "scikit-image!=0.23.0",
 ]
 
--- a/src/allencell_ml_segmenter/_tests/widgets/test_model_download_dialog.py
+++ b/src/allencell_ml_segmenter/_tests/widgets/test_model_download_dialog.py
@@ -7,7 +7,10 @@ from allencell_ml_segmenter._tests.fakes.fake_user_settings import (
     FakeUserSettings,
 )
 from allencell_ml_segmenter.main.experiments_model import ExperimentsModel
-from allencell_ml_segmenter.utils.s3.s3_bucket_constants import PROD_BUCKET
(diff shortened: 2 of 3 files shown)


src/allencell_ml_segmenter/widgets/model_download_dialog.py
```diff
@@ -14,7 +14,7 @@
 from allencell_ml_segmenter.utils.s3.s3_model_bucket import (
     S3ModelBucket,
 )
-from allencell_ml_segmenter.utils.s3.s3_bucket_constants import PROD_BUCKET
+from allencell_ml_segmenter.utils.s3.s3_bucket_constants import STG_BUCKET
```
saeliddp, line 17: Does this have to be manually changed each time we want to update to PROD/STAGE? Could we maybe have an environment variable that we read and default to PROD?

pyproject.toml
```diff
@@ -41,7 +41,7 @@ dependencies = [
     "bioio",
     "tifffile>=2023.4.12",
     "watchdog",
-    "cyto-dl>=0.1.8",
```
hughes036, line 44: why were we so far behind?

src/allencell_ml_segmenter/_tests/widgets/test_model_download_dialog.py
```diff
@@ -41,6 +44,17 @@ def model_download_dialog(
             "adding_headers": {"X-Foo": "Bar"},
         }
     )
+
+    responses.add(
+        **{
+            "method": responses.GET,
+            "url": f"{STG_BUCKET}?list-type=2",
```
hughes036, line 51: Rather that changing this on main, I think we should maintain a STG branch.  Thoughts?

src/allencell_ml_segmenter/widgets/model_download_dialog.py
hughes036: https://github.com/AllenCell/allencell-segmenter-ml/pull/524/files#r1792585526

Yes lets come up with something better.  I fear this will bite us come release time.

PR #530, opened 2024-10-15
Automate tickets created

yrkim98:
Added two workflow files:

- `auto_add_dev_tickets_to_project.yaml` should add all github issues not created with a "Requests" or "Bug Reports" label to the first swimlane of the feature requests board. The new feature templates will make it so that these labels get automatically applied to all feature requests or bug reports submitted
- `auto_add_feature_request_to_project.yaml` should add all github issues without the "Requests" and "Bug Reports" label to the first swimlane of the dev board. This will be for when we create dev tickets to work on. We should not label these as "Requests" or "Bug Reports"

### DIFF
--- a/.github/workflows/auto_add_dev_tickets_to_project.yaml
+++ b/.github/workflows/auto_add_dev_tickets_to_project.yaml
@@ -1,0 +1,19 @@
+name: Auto add dev tickets to project board
+
+on:
+  issues:
+    types: [opened]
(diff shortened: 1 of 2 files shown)


hughes036 (approved):
Cool!

saeliddp (approved):
Nice! Glad there was a clean way to do this in GH actions

PR #538, opened 2024-10-18
Readme language changes

yrkim98:
- Changing more language in the readme based on discussion with graham and gideon
- Fixed emojis in readme
- update "release" to "dev release" in readme > news
- manual version bump to v0.1.16rc3
- latest commit is tagged with v0.1.16rc3

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,11 @@
 [![Test and lint](https://github.com/AllenCell/allencell-segmenter-ml/actions/workflows/test_lint.yaml/badge.svg?branch=main&event=push)](https://github.com/AllenCell/allencell-segmenter-ml/actions/workflows/test_lint.yaml)
 
 ## This version is a release candidate currently undergoing testing and development.
-[Please report any bugs here](https://github.com/AllenCell/allencell-segmenter-ml/issues/new/choose)
+Our team is actively working on this plugin and will have an *official release* of version 0.1.16 with additional features very soon. 
+ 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -3,7 +3,11 @@
 [![Test and lint](https://github.com/AllenCell/allencell-segmenter-ml/actions/workflows/test_lint.yaml/badge.svg?branch=main&event=push)](https://github.com/AllenCell/allencell-segmenter-ml/actions/workflows/test_lint.yaml)
 
 ## This version is a release candidate currently undergoing testing and development.
-[Please report any bugs here](https://github.com/AllenCell/allencell-segmenter-ml/issues/new/choose)
+Our team is actively working on this plugin and will have an *official release* of version 0.1.16 with additional features very soon.
```
saeliddp, line 6: Additional features means we shouldn't call it `0.1.16` right?

yrkim98: that is what graham gave me, but i removed that

saeliddp (approved):

README.md
hughes036: What i said at the meeting is that as soon as we have a fix for the bug report, we should release that as `0.1.16`.  Then when we have iterative training & thresholding, we should release `0.2.0`.  What I really think is that we should release iterative training as `0.2.0`, thresholding as `0.3.0`, do a big push for external QA testing and UX, making the improvements identified, do a `1.0.0` release, and then make a big promotion effort.

PR #540, opened 2024-10-22
Thresholding UI

yrkim98:
Thresholding UI changes, heres what it looks like
image

This is based on Thao's UI available here: https://www.figma.com/design/kiqNACOut8sGePs8tr7W57/Segmenter-ML---napari-plugin?node-id=1634-17590&node-type=frame

**Summary**

- Adding new tab for the thresholding view
- Repurposing `PredictionFileInput` to work for thresholding as well- renamed this to `FileInputWidget`
- Since prediction and thresholding track similar state- created base class `FileInputModel`
- PredictionModel extends `FileInputModel`- adds functions which dispatch prediction-related tasks
- Currently the thresholding view uses `FileInputModel` directly- but planning on changing this so that I have a model specifically for this view that dispatches thresholding-related tasks.
- Getting rid of old pre/post processing code.

**Files changed**

- src/allencell_ml_segmenter/core/file_input_model.py: ABC for  PredictionModel and in the future- PostProcessModel
- src/allencell_ml_segmenter/postprocess/postprocess_view.py: UI for thresholding
- src/allencell_ml_segmenter/prediction/file_input_widget.py: Refactor to work with thresholding as well
- src/allencell_ml_segmenter/prediction/model.py: Now extends `FileInputModel`, an ABC. Only contains functions that dispatch events related to prediction, also removed unused code from old implementation of pre/post processing.
- src/allencell_ml_segmenter/prediction/service.py: Changing type from PredictionModel to FileInputModel
- src/allencell_ml_segmenter/prediction/view.py: Changing type from PredictionModel to FileInputModel
- src/allencell_ml_segmenter/services/prediction_service.py: Changing type from PredictionModel to FileInputModel
- src/allencell_ml_segmenter/styles/thresholding_view.qss: Stylesheet for thresholding


src/allencell_ml_segmenter/prediction/service.py
```diff
@@ -39,13 +33,6 @@ def __init__(self, model: PredictionModel):
     def handle_event(self, event: Event) -> None:
         pass
 
-    def extract_preprocessing_method(self) -> None:
```
hughes036, line 42: Was this dead code, left over from a UI we r emoved?

src/allencell_ml_segmenter/prediction/model.py
```diff
@@ -4,32 +4,16 @@
...
     """
 
     def __init__(self) -> None:
         super().__init__()
-
-        # state related to PredictionFileInput
-        self.config_name: Optional[str] = None
-        self.config_dir: Optional[Path] = None
-        self._input_image_path: Optional[Path] = None
-        self._image_input_channel_index: Optional[int] = None
```
hughes036, line 26: This all went into `FileInputModel` ?

src/allencell_ml_segmenter/main/main_widget.py
```diff
@@ -185,6 +196,7 @@ def _disable_non_prediction_tabs(self, _: Event) -> None:
         """
         self._window_container.setTabEnabled(0, False)
         self._window_container.setTabEnabled(1, False)
+        self._window_container.setTabEnabled(3, True)  # TODO remove
```
hughes036, line 199: why is the same statement on lines 199 and 188 with `TODO remove`?  One needs to be cleaned up?

src/allencell_ml_segmenter/core/file_input_model.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def get_input_mode(self) -> Optional[InputMode]:
+        return self._input_mode
+
+    def set_selected_paths(
+        self, paths: Optional[List[Path]], extract_channels: bool = False
+    ) -> None:
+        self._selected_paths = paths
+        if extract_channels and paths is not None:
+            self.dispatch(Event.ACTION_PREDICTION_EXTRACT_CHANNELS)
```
hughes036, line 62: Is there test cov on these setters which dispatch actions?

src/allencell_ml_segmenter/_tests/prediction/test_model_input_widget.py
```diff
@@ -116,53 +87,12 @@ def test_postprocessing_method(
...
-        == ModelInputWidget.MID_TEXT
-    )
 
     # ACT
     with qtbot.waitSignal(model_input_widget._bottom_postproc_button.toggled):
         model_input_widget._bottom_postproc_button.click()
-
-    # ASSERT
-    assert (
-        model_input_widget._model.get_postprocessing_method()
```
hughes036, line 141: are all the deletes here covering dead code?

src/allencell_ml_segmenter/_tests/prediction/test_model.py
```diff
@@ -56,100 +56,3 @@ def test_output_directory(prediction_model: PredictionModel) -> None:
...
-    # ARRANGE
-    event_under_test: Event = Event.ACTION_PREDICTION_POSTPROCESSING_METHOD
-    subscriber: FakeSubscriber = FakeSubscriber()
-    prediction_model.subscribe(event_under_test, subscriber, subscriber.handle)
-
-    # ACT
-    prediction_model.set_postprocessing_method("example method")
-
-    # ASSERT
-    assert prediction_model.get_postprocessing_method() == "example method"
```
hughes036, line 109: All the deletes here were covering dead code?  Was that code deleted in this pr?

src/allencell_ml_segmenter/prediction/model_input_widget.py
```diff
@@ -309,13 +296,3 @@ def _configure_slots(self) -> None:
         self._bottom_postproc_button.toggled.connect(
             self._bottom_postproc_button_slot
         )
-
-        # connect postprocessing simple threshold slider to slot
-        self._simple_thresh_slider.changed.connect(
```
hughes036, line 314: was this functional before?  I thought we had removed or disabled the post processing UI?

src/allencell_ml_segmenter/postprocess/postprocess_view.py
```diff
@@ -0,0 +1,221 @@
...
+        self._threshold_value_slider.setValue(int(value * 100))
+
+    def _enable_specific_threshold_widgets(self, enabled: bool) -> None:
+        """
+        enable or disable specific value thresholding widgets
+        """
+        self._threshold_value_slider.setEnabled(enabled)
+        self._threshold_value_spinbox.setEnabled(enabled)
+
+    def _enable_apply_button(self) -> None:
```
hughes036, line 201: This is a new widget, which doesn't yet perform thresholding right?  (i see no calls to a model)

hughes036 (approved):
I spotted a a few action dispatches that I think should get tested.  Looks good otherwise.

src/allencell_ml_segmenter/_tests/prediction/test_model_input_widget.py
```diff
@@ -37,10 +37,6 @@ def test_top_radio_button_slot(
     # ASSERT - both input UI elements should be disabled, postprocessing method should be set to TOP_TEXT
```
saeliddp, line 37: nit: if you're getting rid of assertions on postproc method, this reference to postprocessing method in the comment might be confusing

saeliddp: Applies to comments throughout this file

src/allencell_ml_segmenter/core/file_input_model.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def get_max_channels(self) -> Optional[int]:
+        return self._max_channels
+
+    def set_max_channels(self, max: Optional[int]) -> None:
+        self._max_channels = max
+        if max is not None:
+            self.dispatch(Event.ACTION_PREDICTION_MAX_CHANNELS_SET)
```
saeliddp, line 70: Thoughts on converting from home-baked events to slots/signals?

src/allencell_ml_segmenter/core/file_input_model.py
saeliddp, line 62: It seems like `...PREDICTION...` events should not be dispatched from a model that is not specific to prediction

src/allencell_ml_segmenter/prediction/model.py
```diff
@@ -4,32 +4,16 @@
...
+from allencell_ml_segmenter.core.file_input_model import FileInputModel
 
 
-class PredictionInputMode(Enum):
-    FROM_PATH = "from_path"
-    FROM_NAPARI_LAYERS = "from_napari_layers"
-
-
-class PredictionModel(Publisher):
+class PredictionModel(FileInputModel):
```
saeliddp, line 10: What do you think about using model composition instead of inheritance? To me it seems less like `PredictionModel` is a `FileInputModel` and more like `PredictionModel` has a `FileInputModel` because it has a file input widget as part of its UI.

src/allencell_ml_segmenter/prediction/model_input_widget.py
saeliddp: IIRC we removed it from the prediction view but kept the code around so that we could have a starting point for thresholding in the future

src/allencell_ml_segmenter/prediction/service.py
```diff
@@ -17,19 +17,13 @@ class ModelFileService(Subscriber):
     Parses the chosen model file to extract the preprocessing method.
     """
 
-    def __init__(self, model: PredictionModel):
+    def __init__(self, model: FileInputModel):
```
saeliddp, line 20: this should be a `PredictionModel` always, right? If so, I think it's better to be specific in the type annotation

src/allencell_ml_segmenter/styles/thresholding_view.qss
```diff
@@ -0,0 +1,40 @@
+/*
+Global stylesheet
```
saeliddp, line 2: Why does this say global stylesheet? Shouldn't it only apply to thresholding?

saeliddp (approved):
Have some questions/suggestions, but overall looks good!

src/allencell_ml_segmenter/prediction/model.py
hughes036: I think this is probably a good idea, inline with stuff i have read in recent years.

src/allencell_ml_segmenter/prediction/service.py
hughes036: A good piece of evidence for composition over inheritance.

src/allencell_ml_segmenter/prediction/service.py
yrkim98: this is dead code

src/allencell_ml_segmenter/prediction/model.py
yrkim98: config_name and config_dir are dead code.

input_image_path and channel index are used in both thresholding and prediction in `FIleInput` so they have been moved to `FileInputModel`

src/allencell_ml_segmenter/main/main_widget.py
yrkim98: I'll remove this in a subsequent PR before merging this to main. For now it makes thresholding easier to test (the tab is instantly available)

The other line is the behavior we want, but this line makes it so that I can jump to the thresholding tab

src/allencell_ml_segmenter/_tests/prediction/test_model_input_widget.py
yrkim98: this is all dead code

src/allencell_ml_segmenter/_tests/prediction/test_model.py
yrkim98: this is all dead code which we wont use anymore

src/allencell_ml_segmenter/prediction/model_input_widget.py
yrkim98: The thresholding was never functional, but we had a slider in the UI at one point which was removed

src/allencell_ml_segmenter/postprocess/postprocess_view.py
yrkim98: correct. purely UI changes here

src/allencell_ml_segmenter/thresholding/thresholding_service.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def _handle_thresholding_error(self, error: Exception) -> None:
+        show_info("Thresholding failed: " + str(error))
+
+    def _on_threshold_changed(self, _: Event) -> None:
+        # if we havent thresholded yet, keep track of original layers.
+        # need to check this on first threshold change, since user can add images
+        # between finishing prediction and starting thresholding
+        # if they are using images from a directory.
+        original_layers: Optional[OrderedDict[str, np.ndarray]] = (
```
hughes036: nit: dont need this local state (`original_layers`)

src/allencell_ml_segmenter/thresholding/thresholding_service.py
```diff
@@ -0,0 +1,153 @@
...
+            self._thresholding_model.get_original_layers()
+        )
+        if original_layers is None:
+            self._thresholding_model.set_original_layers(
+                self._viewer.get_layers()
+            )
+
+        # Get layers to threshold.
+        # if there are segmentations displayed in the viewer, only threshold those images.
+        layers_to_threshold: OrderedDict[str, np.ndarray] = (
```
hughes036: Doesnt `_thresholding_model` need to be updated to have any layers that might have just been added to the viewer (drag n drop) ?

src/allencell_ml_segmenter/thresholding/thresholding_service.py
yrkim98: this is to satisfy type checking, since the model state is optional, and can be None- we check beforehand

src/allencell_ml_segmenter/thresholding/thresholding_service.py
yrkim98: this should fix this:
https://github.com/AllenCell/allencell-segmenter-ml/blob/0997cd860f166d8292717e0866e119f01496713e/src/allencell_ml_segmenter/thresholding/thresholding_service.py#L67-L69

https://github.com/AllenCell/allencell-segmenter-ml/blob/0997cd860f166d8292717e0866e119f01496713e/src/allencell_ml_segmenter/thresholding/thresholding_service.py#L160-L172

hughes036: Is `self._thresholding_model.get_original_layers()` the layerName -> layer dict that we discussed?

I was imagining the `thresholding_model` encapsulating this functionality, and exposing something like `addLayer(layerName, layer)`.

pyproject.toml
```diff
@@ -33,7 +33,7 @@ dependencies = [
     "npe2>=0.6.2",
     "numpy",
     "hydra-core==1.3.2",
-    "bioio==1.0.1",
+    "bioio==1.1.0",
```
hughes036, line 36: relevant to Thresholding?  Or something that was cherry picked?

src/allencell_ml_segmenter/main/viewer.py
```diff
@@ -23,8 +23,8 @@ def __init__(
         super().__init__()
         self.viewer: napari.Viewer = viewer
 
-    def add_image(self, image: np.ndarray, name: str) -> None:
-        self.viewer.add_image(image, name=name)
+    def add_image(self, image: np.ndarray, **kwargs: Any) -> None:
```
hughes036, line 26: Can you explain this new kwargs usage?

src/allencell_ml_segmenter/main/viewer.py
```diff
@@ -60,8 +61,8 @@ def get_all_shapes(self) -> list[ShapesLayer]:
             if isinstance(l, Shapes)
         ]
 
-    def add_labels(self, data: np.ndarray, name: str) -> None:
-        self.viewer.add_labels(data, name=name)
+    def add_labels(self, data: np.ndarray, **kwargs: Any) -> None:
+        self.viewer.add_labels(data, **kwargs)
```
hughes036, line 77: Is this how you name a layer / labels?

Do we actually need Napari level names if we are implementing our own dict in the data model for managing layers?

src/allencell_ml_segmenter/main/viewer.py
```diff
@@ -167,3 +168,11 @@ def insert_threshold(
             # Thresholding already exists so just update the existing one in the viewer.
             layer_to_insert.data = image
             layer_to_insert.refresh()
+
+    def get_source_path(self, layer: Layer) -> Optional[Path]:
+        if layer.source.path is not None:
+            return Path(layer.source.path)
+        if layer.metadata is not None and "source_path" in layer.metadata:
+            return Path(layer.metadata["source_path"])
```
hughes036, line 179: Can you comment this method?  What is 'source_path' ?  The path of the image on disk?

src/allencell_ml_segmenter/prediction/view.py
```diff
@@ -191,13 +191,16 @@ def showResults(self) -> None:
...
                             raw_np_data,
-                            f"[raw] {data['raw'].name}",
+                            name=f"[raw] {data['raw'].name}",
+                            metadata={"source_path": str(data["raw"])},
                         )
                     if seg_np_data is not None:
                         self._viewer.add_labels(
                             seg_np_data,
                             name=f"[seg] {data['seg'].name}",
+                            metadata={"source_path": str(data["seg"])},
```
hughes036: I see - `name` and `metadata` are the kwargs?

Do we still need to track this info in the actual Napari layers?  Isnt our thresholdModel dict tracking this info now?

src/allencell_ml_segmenter/services/prediction_service.py
```diff
@@ -145,7 +145,6 @@ def build_overrides(self, checkpoint: Path) -> dict[str, Any]:
         # This override is needed to for inference with cyto-dl, because we default to auto otherwise
         overrides["data.batch_size"] = 1
 
-
         input_path: Optional[Path] = (
```
hughes036, line 148: shouldnt our linter be preventing whitespace changes like this?

src/allencell_ml_segmenter/thresholding/thresholding_service.py
```diff
@@ -157,16 +157,26 @@ def _threshold_image(self, image: np.ndarray) -> np.ndarray:
...
+                layer
+                for layer in current_layers
+                if layer.name not in original_layers
+            ]
 
         # refresh layers only if the new layers are not threshold layers (we dont want to track this in original layers state)
         for new_layer in new_layers_added:
             if not new_layer.name.startswith("[threshold]"):
-                self._thresholding_model.set_original_layers(self._viewer.get_layers())
+                self._thresholding_model.set_original_layers(
```
hughes036: Shouldnt linting be enforcing one of these formats, preventing diffs like this?

src/allencell_ml_segmenter/thresholding/thresholding_service.py
```diff
@@ -157,16 +157,26 @@ def _threshold_image(self, image: np.ndarray) -> np.ndarray:
...
 
         # get layers that were added since last thresholding
-        original_layers: Optional[OrderedDict[str, np.ndarray]] = self._thresholding_model.get_original_layers()
+        original_layers: Optional[OrderedDict[str, np.ndarray]] = (
+            self._thresholding_model.get_original_layers()
+        )
         new_layers_added: list[Layer] = current_layers
         if original_layers is not None:
-            new_layers_added: list[Layer] = [layer for layer in current_layers if layer.name not in self._thresholding_model.get_original_layers()]
+            new_layers_added = [
```
hughes036: because of type checking?

src/allencell_ml_segmenter/core/file_input_widget.py
```diff
@@ -214,14 +214,18 @@ def _from_directory_slot(self) -> None:
         self._model.set_input_mode(InputMode.FROM_PATH)
 
     def _update_layer_list(self, event: Optional[NapariEvent] = None) -> None:
-        existing_selection = [
-            self._viewer.get_layers_nonthreshold()[i].name
-            for i in self._image_list.get_checked_rows()
-        ]
+        existing_selection: List[str] = []
+        for i in self._image_list.get_checked_rows():
```
hughes036, line 218: Is `FileInputWidget` the widget that displays a whole list of files in the viewer, and makes you check each one on/off ?

src/allencell_ml_segmenter/main/segmenter_layer.py
```diff
@@ -24,8 +24,9 @@ class ImageLayer(SegmenterLayer):
     path: Optional[Path] = None
     data: Optional[np.ndarray] = None
     source: Source = Source(name="sourcename")
```
saeliddp: I'm still struggling to understand `Source`
1. Why is source inheriting from `SegmenterLayer`--Source is not a Layer right? It's an attribute of a layer
2. Do we actually need to mimic the fact that a real napari layer has a source attribute? It seems to me that we are only using `Source` for `Source.path`, but we already have a `path` attribute on `ImageLayer`. Is there a meaningful distinction between `Source.path` and `path`?

src/allencell_ml_segmenter/core/file_input_widget.py
```diff
@@ -214,14 +214,18 @@ def _from_directory_slot(self) -> None:
...
+            item: Optional[QListWidgetItem] = self._image_list.item(i)
+            if item is not None:
```
hughes036: What is `item` ?

src/allencell_ml_segmenter/core/file_input_widget.py
```diff
@@ -214,14 +214,18 @@ def _from_directory_slot(self) -> None:
         self._model.set_input_mode(InputMode.FROM_PATH)
 
     def _update_layer_list(self, event: Optional[NapariEvent] = None) -> None:
```
hughes036, line 216: Im confused about this being a method of FileInputWidget.  I know there are reasons and that it would be hard to change now.

src/allencell_ml_segmenter/core/file_input_widget.py
```diff
@@ -214,14 +214,18 @@ def _from_directory_slot(self) -> None:
...
         self._image_list.clear()
         self._reset_channel_combobox()
         for layer in self._viewer.get_layers():
-            path_of_layer_image: str = layer.source.path
+            path_of_layer_image: Optional[Path] = self._viewer.get_source_path(
+                layer
+            )
             if path_of_layer_image:
                 self._image_list.add_item(
                     layer.name, set_checked=layer.name in existing_selection
```
hughes036, line 233: It seems weird that `FileInputWidget` relies on `Viewer` for file paths.  I guess it is because when a file is drag/dropped, the Viewer is the first place where that new state can be found?

I think this widget might be better as a single component with the viewer, since they seemed tightly coupled.

src/allencell_ml_segmenter/core/file_input_widget.py
```diff
@@ -230,10 +234,13 @@ def _update_layer_list(self, event: Optional[NapariEvent] = None) -> None:
     def _process_checked_signal(self, row: int, state: Qt.CheckState) -> None:
         if self._model.get_input_mode() == InputMode.FROM_NAPARI_LAYERS:
             selected_indices: List[int] = self._image_list.get_checked_rows()
-            selected_paths: List[Path] = [
-                Path(self._viewer.get_layers_nonthreshold()[i].source.path)
-                for i in selected_indices
-            ]
+            selected_paths: List[Path] = []
+            for i in selected_indices:
```
hughes036, line 240: This looks like a result of the refactoring in `Viewer`.  The previous version of this routine was cleaner imo, what was the gain from refactoring?

src/allencell_ml_segmenter/core/file_input_widget.py
```diff
@@ -257,12 +265,17 @@ def _process_checked_signal(self, row: int, state: Qt.CheckState) -> None:
             elif state == Qt.CheckState.Unchecked:
                 # could have unselected the img we got channels from originally, so need to re-extract
                 # as long as there are still some images selected
-                if len(selected_indices) == 0 and self._include_channel_selection:
+                if (
+                    len(selected_indices) == 0
```
hughes036, line 271: Diffs from here down in this file look like they should be prevented by linting.  Whats up with that?

src/allencell_ml_segmenter/main/viewer.py
saeliddp: It seems like by switching to `**kwargs`, we are thinning the separation between our `Viewer` interface and `napari.Viewer`. I think if this is the route we're going down, we should at minimum note the 'expected' `**kwargs` in a comment. (something like 'Commonly used kwargs are x and y, see napari.viewer.add_labels documentation for a list of all kwargs possible')

src/allencell_ml_segmenter/thresholding/thresholding_service.py
yrkim98: yeah `self._thresholding_model.get_original_layers()` is the dict we discussed.

pyproject.toml
yrkim98: This is a QC issue i found, we require 1.1.0 for the plugin

src/allencell_ml_segmenter/main/viewer.py
yrkim98: yeah the viewer's `add_image()` is a wrapper for napari's `add_image()` function- which has lots of optional kwargs that we can pass to it.
Rather than changing our viewer's method each time we need to add a parameter, its cleaner to pass `**kwargs`, so we can specify any number of keyword arguments without pre-defining them in the viewer method's parameters.

src/allencell_ml_segmenter/main/viewer.py
yrkim98: https://github.com/AllenCell/allencell-segmenter-ml/pull/540#discussion_r1878648437 is why we need the kwargs.

We need napari level names, since that is what's displayed in the viewer. You can pass in name as a `**kwarg` here, or any of the other kwargs the `napari.Viewer.add_labels()` function supports: https://napari.org/dev/api/napari.Viewer.html#napari.Viewer.add_labels

comment is a nice touch. I've added that

src/allencell_ml_segmenter/main/viewer.py
yrkim98: Added comments. `source_path` is the path to the image on disk

src/allencell_ml_segmenter/prediction/view.py
yrkim98: the images added require a name, as that is what's displayed as the layer name in the viewer.

`metadata` is required because images are added in prediction- and we need those images source paths. could be simplified if we simplify the workflow and drop the ability to add any image for thresholding (so that we update `ThresholdingModel` once after prediction finishes)

src/allencell_ml_segmenter/services/prediction_service.py
yrkim98: I added this in a previous commit (the extra whitespace) and the linter caught it and removed it (so there's only one whitespace now.

src/allencell_ml_segmenter/thresholding/thresholding_service.py
yrkim98: not sure why liniting did not catch this previously

src/allencell_ml_segmenter/thresholding/thresholding_service.py
yrkim98: correct

src/allencell_ml_segmenter/core/file_input_widget.py
yrkim98: yes

src/allencell_ml_segmenter/main/segmenter_layer.py
yrkim98: 1) is fixed, should not inherit `SegmenterLayer`
2) we mimic the real source attribute because it is used in tests. namely `prediction/test_view.py::test_show_results` and `prediction/test_view.py::test_show_results_non_empty_folder` when it tries to call showResults() on a real `View`, using a `FakeViewer`

src/allencell_ml_segmenter/core/file_input_widget.py
yrkim98: [`item()` is a `QListWidget` function](https://doc.qt.io/qt-6/qlistwidgetitem.html) which grabs the item at index i.

src/allencell_ml_segmenter/core/file_input_widget.py
yrkim98: This updates the layer list that is displayed in `FileInputWidget`

src/allencell_ml_segmenter/core/file_input_widget.py
yrkim98: > It seems weird that `FileInputWidget` relies on `Viewer` for file paths. I guess it is because when a file is drag/dropped, the Viewer is the first place where that new state can be found?

thats exactly why

Should i create refactor ticket?

src/allencell_ml_segmenter/core/file_input_widget.py
yrkim98: We cant rely on a layer to have a `source.path`, since images added from our plugin will not have that, so we use my new method `viewer.get_source_path()`.

Since `viewer.get_source_path`, i could not do list comprehension like before as we need to do a null check.

src/allencell_ml_segmenter/core/file_input_widget.py
yrkim98: have no idea. Linting has  passed on main and on this branch.
Think it may be from linting on my windows desktop which has a different version of `black`.

yrkim98:
During my latest QC run I realized something: we could store probablity mapping as layer metadata, so even if thresholds are applied and changes the layer data, we have a original probability mapping saved that we can use to update the layer with different thresholds.
This would simplify things a lot- especially with the confusing thresholding layer management we are doing based on names, and would be a lot cleaner, and be less bug prone. (eg. if the user changes the names of the layers in napari- this would be unaffected.)
IMO this is very worth implementing, but not sure I want to make this big of a change 4 days before we are gonna go to ASCB. Since what we have works right now, I'll create a ticket detailing what I want to do so we can do it in the near future.

src/allencell_ml_segmenter/core/file_input_widget.py
hughes036: yeah

src/allencell_ml_segmenter/core/file_input_widget.py
hughes036: please link reactor ticket (discussed in below comment)

src/allencell_ml_segmenter/core/file_input_widget.py
hughes036: but what is the value of it in this context?  Item as a var name tells me nothing.  comment or use a descriptive name?

src/allencell_ml_segmenter/core/file_input_widget.py
hughes036: I think we need a better name for the component.  the current name sounds like its just a single file select element.

hughes036: (not for this pr)

src/allencell_ml_segmenter/prediction/view.py
hughes036: Can you make a ticket for simplifying the workflow, and describe code changes, and link it here?

yrkim98: https://github.com/AllenCell/allencell-segmenter-ml/issues/579

src/allencell_ml_segmenter/core/file_input_widget.py
yrkim98: https://github.com/AllenCell/allencell-segmenter-ml/pull/540/commits/9a753be22f9d2a1ff1536391b2c1906a95a3cb99

src/allencell_ml_segmenter/core/file_input_widget.py
yrkim98: https://github.com/AllenCell/allencell-segmenter-ml/issues/580

src/allencell_ml_segmenter/core/file_input_widget.py
yrkim98: https://github.com/AllenCell/allencell-segmenter-ml/issues/580

hughes036:
> During my latest QC run I realized something: we could store probablity mapping as layer metadata, so even if thresholds are applied and changes the layer data, we have a original probability mapping saved that we can use to update the layer with different thresholds. This would simplify things a lot- especially with the confusing thresholding layer management we are doing based on names, and would be a lot cleaner, and be less bug prone. (eg. if the user changes the names of the layers in napari- this would be unaffected.) IMO this is very worth implementing, but not sure I want to make this big of a change 4 days before we are gonna go to ASCB. Since what we have works right now, I'll create a ticket detailing what I want to do so we can do it in the near future.

Sounds like the right type of thinking.
But:
- will the prob map be displayed as a layer?  If so, how will napari render the image?
- The main concern for retaining this state is to handle a user dragging an image into to viewer after thresholding has already been applied, right?

I still favor imposing more constraints on what the user can do.  I was considering seeing if it is possible to 'lock' the viewer when the thresholding tab is enabled, preventing drag / drop events.  Then they would only be able to apply thresholding changes to images that were added during prediction (automatically, or via drag / drop)

pyproject.toml
```diff
@@ -8,7 +8,7 @@ build-backend = "setuptools.build_meta"
 # https://peps.python.org/pep-0621/
 [project]
 name = "allencell-segmenter-ml"
-version = "1.0.0rc0"
+version = "1.0.0rc1"
```
hughes036: related to plugin changes in this pr?

src/allencell_ml_segmenter/_tests/fakes/fake_viewer.py
```diff
@@ -103,3 +110,41 @@ def subscribe_layers_change_event(
...
+            for layer in self._image_layers.values()
+            if layer.name.startswith("[seg]")
+        ]
+
+    def insert_binary_map_into_layer(
+        self, layer: ImageLayer, img: np.ndarray, seg_layers: bool = False
+    ) -> None:
+        self.threshold_inserted[f"[threshold] {layer.name}"] = img
+
+    def get_layers_nonthreshold(self) -> list[Layer]:
```
hughes036, line 126: Will this be the binary maps, and any images that the user happened to drag in?

hughes036: This method could use some explaining

src/allencell_ml_segmenter/core/event.py
```diff
@@ -61,6 +64,11 @@ class Event(Enum):
     ACTION_CURATION_SEG1_THREAD_ERROR = "curation_seg1_thread_error"
     ACTION_CURATION_RAW_THREAD_ERROR = "curation_raw_thread_error"
 
+    # Thresholding events
+    ACTION_EXECUTE_THRESHOLDING = "execute_thresholding"
+    ACTION_SAVE_THRESHOLDING_IMAGES = "save_thresholding_images"
+    ACTION_THRESHOLDING_DISABLED = "disable_thresholding"
```
hughes036, line 70: As discussed in a ticket, i think it would be better if our ux design didnt require `ACTION_THRESHOLDING_DISABLED`

src/allencell_ml_segmenter/core/file_input_widget.py
```diff
@@ -146,33 +148,35 @@ def __init__(
         frame_layout.addLayout(horiz_layout)
 
         grid_layout: QGridLayout = QGridLayout()
+        if include_channel_selection:
+            image_input_label: LabelWithHint = LabelWithHint(
```
hughes036, line 152: This is part of generalizing the component for thresholding use?

src/allencell_ml_segmenter/core/file_input_widget.py
```diff
@@ -256,49 +267,60 @@ def _process_checked_signal(self, row: int, state: Qt.CheckState) -> None:
             elif state == Qt.CheckState.Unchecked:
                 # could have unselected the img we got channels from originally, so need to re-extract
                 # as long as there are still some images selected
-                if len(selected_indices) == 0:
-                    self._service.stop_channel_extraction()  # stop so combobox doesn't reset after thread is finished
+                if (
```
hughes036, line 270: formatting looks weird - did the right version of black produce this?

src/allencell_ml_segmenter/core/file_input_widget.py
```diff
@@ -256,49 +267,60 @@ def _process_checked_signal(self, row: int, state: Qt.CheckState) -> None:
...
+                    len(selected_indices) == 0
+                    and self._include_channel_selection
+                ):
+                    if self._service:
```
hughes036, line 274: Is `self._service` being truthy a heuristic for something?

hughes036: is it about this being a prediction vs thresholding use case?

src/allencell_ml_segmenter/core/file_input_widget.py
```diff
@@ -256,49 +267,60 @@ def _process_checked_signal(self, row: int, state: Qt.CheckState) -> None:
...
+                    extract_channels=len(selected_paths) > 0
+                    and self._include_channel_selection,
                 )
 
     def _reset_channel_combobox(self) -> None:
-        self._channel_select_dropdown.clear()
-        self._channel_select_dropdown.setPlaceholderText("")
-        self._channel_select_dropdown.setCurrentIndex(-1)
-        self._channel_select_dropdown.setEnabled(False)
+        if self._include_channel_selection:
```
hughes036, line 285: is this condition about this being a prediction vs thresholding use case?

Needing this all over says to me that the abstraction need some work.

src/allencell_ml_segmenter/main/main_model.py
```diff
@@ -32,6 +32,9 @@ def __init__(self) -> None:
         self._current_view: Optional[MainWindow] = None
         self._is_new_model: bool = False
         self.signals: MainModelSignals = MainModelSignals()
+        self._predictions_in_viewer: bool = (
+            False  # Tracks whether predictions are displayed in the viewer
```
hughes036, line 36: How is this info used?  Is this something about clearing images when the tab changes?

src/allencell_ml_segmenter/thresholding/thresholding_view.py
```diff
@@ -0,0 +1,349 @@
...
+            self._viewer.get_all_layers_containing_prob_map()
+        ):
+            if idx in self._thresholding_model.get_selected_idx():
+                selected_images_viewer.append(layer.metadata["source_path"])
+        self._thresholding_model.set_selected_paths(selected_images_viewer)
+
+        if output_dir is not None and self._check_able_to_threshold():
+            # progress tracker is tracking number of images saved to the thresholding folder
+            progress_tracker: PredictionFolderProgressTracker = (
+                PredictionFolderProgressTracker(
```
hughes036: `PredictionFolderProgressTracker` seems like this shoudl have a more general name now that we also use it in thresholding.

src/allencell_ml_segmenter/thresholding/thresholding_view.py
```diff
@@ -0,0 +1,349 @@
...
+        )
+        selected_images_viewer: list[Path] = []
+        for idx, layer in enumerate(
+            self._viewer.get_all_layers_containing_prob_map()
+        ):
+            if idx in self._thresholding_model.get_selected_idx():
+                selected_images_viewer.append(layer.metadata["source_path"])
+        self._thresholding_model.set_selected_paths(selected_images_viewer)
+
+        if output_dir is not None and self._check_able_to_threshold():
```
hughes036, line 319: if `_check_able_to_threshold` is false, what feedback does the user get?

hughes036: IMO it would be better if the save button was only enabled when the `_check_able_to_threshold` conditions are true.

src/allencell_ml_segmenter/main/viewer.py
```diff
@@ -105,3 +129,79 @@ def _get_layer_by_name(self, name: str) -> Optional[Layer]:
...
+            for layer in self.get_layers()
+            if "prob_map" in layer.metadata
+        ]
+
+    def clear_binary_map_from_layer(self, layer: Layer) -> None:
+        """
+        We need to keep the layer because it contains the segmentation's probability map.
+        So, clear out the binary map by setting it to all zeros
+        """
+        if "threshold_applied" in layer.metadata:
```
hughes036, line 197: These metadata keys are hardcoded in a few places I imagine? If so extracting constants would good.

src/allencell_ml_segmenter/main/viewer.py
```diff
@@ -94,6 +107,17 @@ def contains_layer(self, name: str) -> bool:
...
 
+    def get_layers_nonthreshold(self) -> list[Layer]:
+        """
+        Get only layers which are not segmentation layers from the viewer.
+        These are the layers that do not start with [threshold].
+        """
+        return [
+            l
+            for l in self.viewer.layers
+            if not l.name.startswith("[threshold]")
```
hughes036, line 118: `"[threshold]"` is repeated a lot - should be a constant.

src/allencell_ml_segmenter/_tests/fakes/fake_viewer.py
```diff
@@ -137,3 +137,14 @@ def get_source_path(self, layer: Layer) -> Optional[Path]:
             return Path(layer.metadata["source_path"])
 
         return None
+
+    def get_all_layers_containing_prob_map(self) -> list[Layer]:
+        return [
+            layer
+            for layer in self.get_all_images()
+            if getattr(layer, "metadata", None)
```
saeliddp, line 145: (I forget how python deals with this) would `if layer.metadata and...` be equivalent and satisfy mypy? if so, it's prob a bit more readable

src/allencell_ml_segmenter/core/file_input_widget.py
saeliddp: I am also wondering about this

src/allencell_ml_segmenter/main/main_widget.py
```diff
@@ -97,7 +97,6 @@ def __init__(
             self._model,
         )
 
-        self._thresholding_file_input_model: FileInputModel = FileInputModel()
```
saeliddp: Nice! Makes the main widget more interpretable

src/allencell_ml_segmenter/main/viewer.py
```diff
@@ -141,45 +141,31 @@ def get_seg_layers(self) -> list[Layer]:
...
-            if seg_layer_og:
-                layerlist.remove(seg_layer_og)
-        else:
-            # Thresholding already exists so just update the existing one in the viewer.
-            layer_to_insert.data = image
-            layer_to_insert.refresh()
+        # if threshold has not been previously applied, update name
+        if (
+            "threshold_applied" not in layer.metadata
+            or not layer.metadata["threshold_applied"]
```
saeliddp, line 163: If the threshold has been previously applied, wouldn't the layer's name be `f"[threshold] {layer.name}"`. It might be more readable to check if the layer's name is already correct, and update it otherwise

saeliddp: Or actually, is there anything wrong with just updating it regardless?

src/allencell_ml_segmenter/thresholding/thresholding_model.py
```diff
@@ -75,41 +75,32 @@ def set_threshold_enabled(self, enabled: bool) -> None:
...
-        return self._original_layers_in_viewer
+    def set_thresholding_layers(self, layers: list[Layer]) -> None:
+        self._thresholding_layers = layers
+
+    def get_thresholding_layers(self) -> list[Layer]:
+        return self._thresholding_layers
+
+    def set_selected_idx(self, selected_idx: list[int]) -> None:
+        self._selected_idx = selected_idx
+        self.dispatch(Event.ACTION_EXECUTE_THRESHOLDING)
```
saeliddp, line 103: Maybe only dispatch if idx changes?

saeliddp (approved):
It's hard for me to keep the entire scope of this change in mind while reviewing, but I left a few comments on some small things I noticed--I assume we are at a point anyway where you and chris have iterated on general design multiple times

src/allencell_ml_segmenter/thresholding/thresholding_model.py
yrkim98: idx will always change on set_selected_idx calls, since it is called when someone selects or unselects an image

src/allencell_ml_segmenter/main/viewer.py
yrkim98: need to do it this way since if threshold_applied == false, I need to drop the `[threshold]` from the name

src/allencell_ml_segmenter/_tests/fakes/fake_viewer.py
yrkim98: this is only all images that are in the viewer

src/allencell_ml_segmenter/core/file_input_widget.py
yrkim98: it is.

src/allencell_ml_segmenter/core/file_input_widget.py
yrkim98: yes this is the correct version of black. I believe due to the extra condition added right below this line

src/allencell_ml_segmenter/core/file_input_widget.py
yrkim98: not truthy, just if it exists and is not None.

src/allencell_ml_segmenter/core/file_input_widget.py
yrkim98: https://github.com/AllenCell/allencell-segmenter-ml/issues/591 created for this

src/allencell_ml_segmenter/main/main_model.py
yrkim98: this is set to True in prediction to tell us that there are plugin generated predictions in the viewer.

src/allencell_ml_segmenter/thresholding/thresholding_view.py
yrkim98: https://github.com/AllenCell/allencell-segmenter-ml/pull/540/commits/098f0759c11ad3efeff941be0e154d2575da6fb8

src/allencell_ml_segmenter/thresholding/thresholding_view.py
yrkim98: There is a popup dialog in napari that shows the message

yrkim98: https://github.com/AllenCell/allencell-segmenter-ml/issues/592

src/allencell_ml_segmenter/_tests/fakes/fake_viewer.py
yrkim98: the getattr is actually needed here to satisfy mypy

PR #542, opened 2024-10-29
Iterative training hotfix

yrkim98:
- This changes regular prediction runs to override the correct ckpt_path key in the cyto-dl config
- Also explicitly turns off iterative training- so we dont have to rely on cyto-dl configs to have iterative training turned off by default
- Update tests

### DIFF
--- a/src/allencell_ml_segmenter/utils/cyto_overrides_manager.py
+++ b/src/allencell_ml_segmenter/utils/cyto_overrides_manager.py
@@ -62,6 +62,9 @@ class CytoDLOverridesManager:
                 )
             # Filters/Model Size (required if starting new model)
             overrides_dict["model._aux.filters"] = model_size.value
+            # needed to specify no weights used
+            overrides_dict["checkpoint.weights_only"] = False
+            overrides_dict["checkpoint.strict"] = True
 
         # Hardware overrides (required)
         if CUDAUtils.cuda_available():
--- a/src/allencell_ml_segmenter/utils/cyto_overrides_manager.py
+++ b/src/allencell_ml_segmenter/utils/cyto_overrides_manager.py
@@ -62,9 +62,6 @@ class CytoDLOverridesManager:
                 )
             # Filters/Model Size (required if starting new model)
             overrides_dict["model._aux.filters"] = model_size.value
-            # needed to specify no weights used
-            overrides_dict["checkpoint.weights_only"] = False
-            overrides_dict["checkpoint.strict"] = True
 
         # Hardware overrides (required)
         if CUDAUtils.cuda_available():
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,7 +41,7 @@ dependencies = [
     "bioio",
     "tifffile>=2023.4.12",
     "watchdog",
-    "cyto-dl>=0.3.0",
+    "cyto-dl>=0.4.1",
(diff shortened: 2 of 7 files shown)


src/allencell_ml_segmenter/utils/cyto_overrides_manager.py
```diff
@@ -62,6 +62,9 @@ def get_training_overrides(
                 )
             # Filters/Model Size (required if starting new model)
             overrides_dict["model._aux.filters"] = model_size.value
+            # needed to specify no weights used
+            overrides_dict["checkpoint.weights_only"] = False
```
hughes036, line 66: how does `checkpoint.weights_only` get overidden?

hughes036: was that already implemented ?

hughes036: Does the config itself have a default?  It seems like if we want to default to False , that would be the ideal place to do it.

saeliddp: Looks like that happens here: [in the overrides manager class](https://github.com/AllenCell/allencell-segmenter-ml/blob/5676077d5f81e3c519e18bcea9bdf5c98dfd8938/src/allencell_ml_segmenter/utils/cyto_overrides_manager.py#L42)

saeliddp (approved):

src/allencell_ml_segmenter/utils/cyto_overrides_manager.py
hughes036: Is there test cov for `weights_only` ?

hughes036: in `test_get_training_overrides_iterative_training` i see as asserts True on `weights_only` but i dont see anything setting `is_using_existing_model` in the ARRANGE section.

yrkim98: `checkpoint.weights_only` should be True for iterative training runs and False for regular training runs.

The config's defaults are `checkpoint.weights_only = False` and `checkopint.strict = True` so we shouldnt need to explicitly set here, just wanted to ensure it does get set for non iterative training runs. I will remove this.

yrkim98: > in `test_get_training_overrides_iterative_training` i see as asserts True on `weights_only` but i dont see anything setting `is_using_existing_model` in the ARRANGE section.

In this test we are setting `model.set_is_using_existing_model(True)` when arranging on line 117.

hughes036: Thanks for pointing out 117, i missed it .  lgtm.

hughes036 (approved):

src/allencell_ml_segmenter/utils/cyto_overrides_manager.py
```diff
@@ -62,6 +62,9 @@ def get_training_overrides(
...
+            overrides_dict["checkpoint.strict"] = True
```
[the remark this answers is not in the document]
benjijamorris: strict means that every weight's name and shape in the checkpoint file has to match the model names/shapes that it's being loaded into

PR #545, opened 2024-10-30
Update prediction to use cyto-dl==0.4.1

yrkim98:
- src/allencell_ml_segmenter/prediction/model.py: Changed where we grab prediction outputs to display: now saved in `seg`
- src/allencell_ml_segmenter/services/prediction_service.py: Channel override changed to `input_channel`, also checking for `None` explicitly so a falsy value (like `channel==0`) still overrides the config.
- Test updates to reflect these changes


src/allencell_ml_segmenter/services/prediction_service.py
```diff
@@ -157,10 +157,8 @@ def build_overrides(self, checkpoint: Path) -> dict[str, Any]:
         channel: Optional[int] = (
             self._prediction_model.get_image_input_channel_index()
         )
-        if channel:
-            overrides["data.transforms.predict.transforms[1].reader[0].C"] = (
-                channel
-            )
+        if channel is not None:
```
saeliddp, line 160: I feel like I always run into bugs like this--almost wish python didn't do 'truthy'/'falsy'

saeliddp (approved):

src/allencell_ml_segmenter/services/prediction_service.py
hughes036: js would shine here..

src/allencell_ml_segmenter/prediction/model.py
```diff
@@ -81,7 +81,7 @@ def get_output_seg_directory(self) -> Optional[Path]:
         Gets path to where segmentation predictions are stored.
         """
         return (
-            self._output_directory / "target"
+            self._output_directory / "seg"
```
hughes036, line 84: test cov?

hughes036 (approved):

src/allencell_ml_segmenter/prediction/model.py
hughes036: spoke in person

PR #547, opened 2024-11-05
Separate file input model from prediction model

yrkim98:
**InitalThoughts**
Initially in https://github.com/AllenCell/allencell-segmenter-ml/commit/c2fecd2d561e3e079b90bad52051478c844b3fd9 I decided to init FileInputModel within PredictionModel like suggested
but that was a bad idea- had to access nested functions multiple layers down in our Services/Views and tests.
https://github.com/AllenCell/allencell-segmenter-ml/blob/c5355369a4ad6ea8cdb480bd2cc69933da5a4508/src/allencell_ml_segmenter/services/prediction_service.py#L140-L142

**Second Pass**
In https://github.com/AllenCell/allencell-segmenter-ml/commit/2b2adba69721b8d71b19758539b7b56bb8aef5d3 I changed this so we now init `FileInputModel` when we need to use it along with our other models in main_widget.py. This way, `FileInputModel` and `PredictionModel` are completely separated. `FileInputModel` now tracks all state related to image inputs / outputs, and `PredictionModel` is responsible for dispatching events and tracking prediction state (only tracking `total_num_images` currently- the total number of images predicted). 

https://github.com/AllenCell/allencell-segmenter-ml/blob/2b2adba69721b8d71b19758539b7b56bb8aef5d3/src/allencell_ml_segmenter/main/main_widget.py#L87
This is then passed to the views and services that need it:
https://github.com/AllenCell/allencell-segmenter-ml/blob/2b2adba69721b8d71b19758539b7b56bb8aef5d3/src/allencell_ml_segmenter/main/main_widget.py#L106-L110

and this makes testing it a lot cleaner too
https://github.com/AllenCell/allencell-segmenter-ml/blob/2b2adba69721b8d71b19758539b7b56bb8aef5d3/src/allencell_ml_segmenter/_tests/services/test_prediction_service.py#L103-L126

### DIFF
--- a/src/allencell_ml_segmenter/_tests/prediction/test_file_input_widget.py
+++ b/src/allencell_ml_segmenter/_tests/prediction/test_file_input_widget.py
@@ -39,7 +39,7 @@ def file_input_widget(
 def test_top_radio_button_slot(
     qtbot: QtBot,
     file_input_widget: FileInputWidget,
-        file_input_model: FileInputModel,
+    file_input_model: FileInputModel,
 ) -> None:
     """
     Test the _top_radio_button_slot method of FileInputWidget.
@@ -61,7 +61,7 @@ def test_top_radio_button_slot(
 def test_bottom_radio_button_slot(
     qtbot: QtBot,
     file_input_widget: FileInputWidget,
-        file_input_model: FileInputModel,
+    file_input_model: FileInputModel,
 ) -> None:
     """
     Test the _bottom_radio_button_slot method of FileInputWidget.


src/allencell_ml_segmenter/main/main_widget.py
```diff
@@ -83,7 +84,7 @@ def __init__(
         self._training_model: TrainingModel = TrainingModel(
             main_model=self._model, experiments_model=self._experiments_model
         )
-
+        self._file_input_model: FileInputModel = FileInputModel()
```
saeliddp, line 87: Why are we initializing the `FileInputModel` here for prediction, but not for postprocessing?

src/allencell_ml_segmenter/_tests/prediction/test_file_input_widget.py
```diff
@@ -11,14 +12,18 @@
...
+    FileInputModel,
+)
 
 MOCK_PATH: str = "/path/to/file"
 
 
 @pytest.fixture
-def prediction_model(qtbot: QtBot) -> PredictionModel:
-    return PredictionModel()
+def prediction_model(qtbot: QtBot) -> FileInputModel:
```
saeliddp, line 25: This is confusing that the fixture called `prediction_model` returns a `FileInputModel`

saeliddp: there's also more 'prediction' wording in this file that should prob be replaced since the widget isn't specific to prediction anymore

saeliddp:
> but that was a bad idea- had to access nested functions multiple layers down in our Services/Views and tests.

It's verbose, yeah, but I don't think it's necessarily bad. 

In my mind, the prediction model represents the state of the whole prediction tab/view. And within the prediction view, we display a file input widget which is complex (and reusable) enough to have it's own model. So to me, it makes the most sense to have `FileInputModel` as a member of `PredictionModel`, just as `FileInputWidget` is a member of `PredictionView`.

Completely separating the `FileInputModel` from the `PredictionModel` and initializing them as top level entities in `MainWidget` seems unconventional to me since `FileInputWidget` (and thus the model) are not present as tabs in the `MainWidget`.

If the verbosity of `prediction_model.file_input_model` is too bothersome (and I can definitely see it becoming ridiculous if we ended up with > 2 levels of model nesting), then it's always possible to expose the file input model methods through proxy methods on prediction model. That option is still a little tedious though to write all of those methods.

At the end of the day, though, these are stylistic, not functional preferences, so it's possible I'm just being a bit of a stickler :)

hughes036:
@yrkim98 @saeliddp I think you guys articulated both sides of a valid trade off.  @yrkim98 when you do composition instead of inheritance, the idea is to encapsulate the model you are composing.  So you would not want to just expose it via a getter, the composing model would proxy calls needed to the inner model.

I agree with you @yrkim98 about what you have dine here producing really clean tests.  @saeliddp I think this is basically a trade off we have made from the jump, and as a result a lot of details have leaked in to the top level.  It has been a learning experience.  Im not sure that I would go with this same design in the next pyqt app.  But in context i think Brian's choice here is a pretty good one.

hughes036 (approved):

saeliddp (approved):

PR #550, opened 2024-11-08
Enable testing/building on 3.9 in ci

yrkim98:
Testing on python 3.9 in ci.

using this to build on python 3.9 in our `publish.yaml` workflow, and we can add any other supported versions here in the future.
https://github.com/AllenCell/allencell-segmenter-ml/blob/585476cc9b4d08c03c892109173581fb3d95f79e/.github/workflows/publish.yaml#L32-L34

`test_lint_pr.yaml` and `test_lint_push.yaml` already had the matrix strategy, so just adding 3.9 as one of the versions to test on.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,13 +11,8 @@ name = "allencell-segmenter-ml"
 version = "0.1.16"
 description = "A plugin to leverage ML segmentation in napari"
 readme = "README.md"
-requires-python = "<3.11,>=3.9"
+requires-python = ">=3.9,<3.11"
 license = { file = "LICENSE" }
-authors = [
-    { email = "brian.kim@alleninstitute.org", name = "Brian Kim" },
-    { email = "chrishu@alleninstitute.org", name = "Chris Hughes" },
-    { email = "daniel.saelid@alleninstitute.org", name = "Daniel Saelid" },
-]
 
 classifiers = [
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yaml
```diff
@@ -29,6 +29,9 @@ jobs:
     permissions:
       id-token: write
       contents: read
+    strategy:
+      matrix:
+        python-version: [ "3.9", "3.10" ]
```
saeliddp, line 34: Is this common practice to upload multiple times to pypi with the same version number but different versions of python used to build?

amilworks: I think multiple Python versions can be associated with one version number, as long as they are packaged as separate wheels or distributions that specify compatibility for different Python versions.

.github/workflows/publish.yaml
```diff
@@ -29,6 +29,9 @@ jobs:
     permissions:
       id-token: write
       contents: read
+    strategy:
+      matrix:
```
amilworks, line 33: Since we are using a matrix setup for testing multiple versions, do you think we should add the `fail-fast` option? 

```yaml
strategy:
  fail-fast: false
```

amilworks (approved):

saeliddp (approved):

PR #555, opened 2024-11-13
make napari and pyqt install optional, and update readme

yrkim98:
## Purpose
This makes `napari` and `pyqt` optional. Bundled app users will not need to install these. When installing in an environment, we now prompt the user to run `pip install napari-segmenter-ml[napari]` to grab everything including pyqt5 and napari

## Changes
Makes it nicer to install plugin on bundled app (currently the double napari install is a source of slowness)

## Testing
tested locally

## How to review
`pyproject.toml`: created optional dependency to install napari and pyqt5
`readme.md`: updated w/ new instructions

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+*   @hughes036 @yrkim98 @saeliddp
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,5 +1,1 @@
 blank_issues_enabled: false
-contact_links:
-  - name: Allen Cell Segmenter Support
-    url: https://github.com/AllenCell/allencell-segmenter-ml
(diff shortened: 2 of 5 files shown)


hughes036 (approved):

saeliddp:
Looks like tests are failing

pyproject.toml
```diff
@@ -39,6 +38,7 @@ dependencies = [
     "watchdog",
     "cyto-dl>=0.4.2",
     "scikit-image!=0.23.0",
+    "magicgui"
```
saeliddp, line 41: Why are we adding this dependency?

saeliddp (approved):

pyproject.toml
yrkim98: this should have been under `test_lint`- changed https://github.com/AllenCell/allencell-segmenter-ml/pull/555/commits/d44403820e1788213828147f3e71f2f7b8abc66f

PR #573, opened 2024-11-27
Feature/thresholding live images

yrkim98:
## Purpose
This PR adds thresholding to the plugin.
here is a demo of the feature

https://github.com/user-attachments/assets/232d3358-157f-4c0a-a5c3-82b90058effe

- When a new threshold value/method is selected, images in the napari viewer have the threshold applied and displayed with the new value.
- If coming from prediction on images from the viewer, plugin knows to only threshold segmentations produced during prediction, and updates the correct layer with the thresholded image.
- Can threshold images from a directory of images

If there's any issues with the PR or you need my help to explain anything, just let me know!

## Changes
There's a lot of changes, but the meat of the changes are in the model, view, and service files under `allencell_ml_segmenter/thresholding`. These follow patterns similar to what we've made in other parts of the plugin:
- `src/allencell_ml_segmenter/thresholding/thresholding_view.py`: Contains the UI which you guys have already seen. UI elements now update the `ThresholdingModel` and the updates to the model dispatch events which implement the features described above. I am using the `View` pattern of using `doWork` to start a task off the main thread to keep the UI responsive.
- `src/allencell_ml_segmenter/thresholding/thresholding_model.py`: Handles state changes from the View, and dispatches events to have the ThresholdingService react to these state changes
- `src/allencell_ml_segmenter/thresholding/thresholding_service.py`: Implements functions which threshold images from the viewer, thresholds images from a directory, and saves images,

The following changes are the rest of the relevant changes for this PR:
- `src/allencell_ml_segmenter/core/file_input_widget.py`: Thresholding UI uses this widget, but does not use channel selection, so made this edit to make it optional. Also changed update_layer_list behavior so that it maintains previous selections, which we need for thresholding updates, and is better for the UI (if a user adds a new image, the previous selections are kept, and the list refreshes with the new image)
- `src/allencell_ml_segmenter/core/file_input_model.py`: is a method that returns a list of images, from the viewer or directory, that the `file_input_widget` has selected
- `src/allencell_ml_segmenter/main/viewer.py`: Added ability to insert a thresholded segmentation image. Has logic for creating the layer in the correct position if needed, or if the layer already exists updates the existing layer with the image.
- `src/allencell_ml_segmenter/main/main_model.py`: adds state to track if there are predictions in viewer displayed from the prediction part of the plugin, which thresholding uses to update the correct layers if needed.
- `src/allencell_ml_segmenter/styles/thresholding_view.qss`: adds styling for disabled ui elements

The rest of the changes are less relevant:
There are downstream changes from the above changes, like test fixes or updating interfaces.

I changed `ExperimentsModel`, which admittedly I should have separated from this PR. Was getting a bug that wasnt letting me run my debugger and thought this would be a small change but it had some downstream changes that made this change bigger than I would have liked.
Main thing here is that when we added typing to `_get_exp_path()` we made it always return a Path, but there are places in the plugin where `None` is a valid state. This ended up throwing some errors when the plugin was started.

## Testing
I tested by installing on the windows desktop and running everything.
Also have 3 new tests added for this:
- `src/allencell_ml_segmenter/_tests/thresholding/test_thresholding_model.py` is a pretty small test file, which just ensures that the correct events are dispatched as we expect when certain state is changed
- `src/allencell_ml_segmenter/_tests/thresholding/test_thresholding_service.py` tests the service through a Faked viewer. I set state in the model in this test and check to see if the service runs the correct function for that state change and that the viewer is updated correctly.
- `src/allencell_ml_segmenter/_tests/thresholding/test_thresholding_view.py` tests to see if changes to the view propagates correctly to the rest of the ui and the model

## How to review
Apologies for the large pr
I would start with the Model, View, and Service in `allencell_ml_segmenter/thresholding` and their tests. then move on to the 5 relevant files I mentioned above, which should cover most of the thresholding related changes.

The rest of the changes are smaller changes changes for fixing broken tests, the ExperimentModel change, or downstream changes from the above updates.

## Note
We'll need to change the cyto-dl version pin with some changes needed for thresholding. I'll have a PR out over there with the changes needed for thresholding and do a 0.4.5 release for that before I merge this to main.


hughes036:
> We'll need to change the cyto-dl version pin with some changes needed for thresholding. I'll have a PR out over there with the changes needed for thresholding and do a 0.4.5 release for that before I merge this to main.

Was all your testing done using the new cyto-dl version?

hughes036:
Great job on this pr description!

pyproject.toml
```diff
@@ -41,7 +41,7 @@ dependencies = [
     "bioio",
     "tifffile>=2023.4.12",
     "watchdog",
-    "cyto-dl>=0.1.8",
+    "cyto-dl>=0.4.4",
```
hughes036, line 44: was all your testing done on 0.4.4 ?  Will it need to be repeated on 0.4.5?

src/allencell_ml_segmenter/thresholding/thresholding_service.py
```diff
@@ -0,0 +1,144 @@
...
+
+    def _handle_thresholding_error(self, error: Exception) -> None:
+        show_info("Thresholding failed: " + str(error))
+
+    def _on_threshold_changed(self, _: Event) -> None:
+        # if we havent thresholded yet, keep track of original layers.
+        if not self._original_layers:
+            self._original_layers = self._viewer.get_layers()
+
+        layers_to_threshold: list[Layer] = self._original_layers
```
hughes036: Every thresholding changes, `layers_to_threshold` should be set to `self._original_layers`? 

Can self._viewer.get_layers() change in between thresholding changes?

Asking because it looks like you could simplify by just set `layers_to_threshold = self._viewer.get_layers()`

src/allencell_ml_segmenter/thresholding/thresholding_service.py
```diff
@@ -0,0 +1,144 @@
...
+        seg_layers: bool = False
+        if self._main_model.are_predictions_in_viewer():
```
hughes036: Im getting lost reading the rest of this routine

src/allencell_ml_segmenter/thresholding/thresholding_service.py
```diff
@@ -0,0 +1,144 @@
...
+            self._task_executor.exec(
+                task=thresholding_task,
+                # lambda functions capture variables by reference so need to pass layer as a default argument
+                on_return=on_return,
+                on_error=self._handle_thresholding_error,
+            )
+
+    def _save_thresholded_images(self, _: Event) -> None:
+        images_to_threshold: list[Path] = (
+            self._file_input_model.get_input_files_as_list()
```
hughes036, line 122: does this return the images in the viewer, or in the selected directory?

src/allencell_ml_segmenter/thresholding/thresholding_service.py
```diff
@@ -0,0 +1,144 @@
...
+
+    def _save_thresh_image(
+        self, image: np.ndarray, original_image_name: str
+    ) -> None:
+        output_directory: Optional[Path] = (
+            self._file_input_model.get_output_directory()
+        )
+        if output_directory is not None:
+            new_image_path: Path = (
+                output_directory / f"threshold_{original_image_name}"
```
hughes036, line 145: will this allow them to overwrite existing thresholded images (if the save multiple times)?

src/allencell_ml_segmenter/thresholding/thresholding_view.py
```diff
@@ -0,0 +1,343 @@
...
+
+        self._threshold_value_spinbox: QSpinBox = QSpinBox()
+        self._threshold_value_spinbox.setRange(
+            THRESHOLD_RANGE[0], THRESHOLD_RANGE[1]
+        )
+        self._threshold_value_spinbox.setSingleStep(1)
+
+        # set default value
+        self._threshold_value_slider.setValue(
+            self._thresholding_model.get_thresholding_value()
```
hughes036, line 143: The model will just have defaults?  Or potentially have values from previous thresholding attempts?

src/allencell_ml_segmenter/training/image_selection_widget.py
```diff
@@ -134,8 +134,9 @@ def __init__(
         self._model.signals.num_channels_set.connect(self._update_channels)
 
     def set_inputs_csv(self, event: Optional[Event] = None) -> None:
-        if self._experiments_model.get_csv_path() is not None:
-            csv_path = self._experiments_model.get_csv_path() / "train.csv"
+        csv_path: Optional[Path] = self._experiments_model.get_csv_path()
+        if csv_path is not None:
```
hughes036, line 138: How does this change relate to the pr?

src/allencell_ml_segmenter/thresholding/thresholding_view.py
```diff
@@ -0,0 +1,343 @@
...
+                PredictionFolderProgressTracker(
+                    output_dir,
+                    len(self._file_input_model.get_input_files_as_list()),
+                )
+            )
+
+            self.startLongTaskWithProgressBar(progress_tracker)
+
+    def doWork(self) -> None:
+        self._thresholding_model.dispatch_save_thresholded_images()
```
hughes036, line 326: for an image dir, is this when thresholding is applied, or just when disk io happens?

src/allencell_ml_segmenter/thresholding/thresholding_view.py
```diff
@@ -0,0 +1,343 @@
...
+
+    def _save_thresholded_images(self) -> None:
+        output_dir: Optional[Path] = (
+            self._file_input_model.get_output_directory()
+        )
+        if output_dir is not None and self._check_able_to_threshold():
+            progress_tracker: PredictionFolderProgressTracker = (
+                PredictionFolderProgressTracker(
+                    output_dir,
+                    len(self._file_input_model.get_input_files_as_list()),
```
hughes036, line 319: What change will this be watching for (tracking)?  That would be a good comment.

src/allencell_ml_segmenter/thresholding/thresholding_view.py
```diff
@@ -0,0 +1,343 @@
...
+        )
+
+        self._apply_save_button.setEnabled(
+            self._specific_value_radio_button.isChecked()
+            or self._autothreshold_radio_button.isChecked()
+        )
+
+    def _check_able_to_threshold(self) -> bool:
+        able_to_threshold: bool = True
+        # Check to see if output directory is selected
```
hughes036, line 276: There is a lot of logic in this, does it have test cov?

src/allencell_ml_segmenter/thresholding/thresholding_service.py
```diff
@@ -0,0 +1,144 @@
...
+            self,
+            self._save_thresholded_images,
+        )
+
+    def _handle_thresholding_error(self, error: Exception) -> None:
+        show_info("Thresholding failed: " + str(error))
+
+    def _on_threshold_changed(self, _: Event) -> None:
+        # if we havent thresholded yet, keep track of original layers.
+        if not self._original_layers:
```
hughes036: Spoke in person about moving `self._original_layers` into `thresholdingModel`.

hughes036: We seem to have an issue of `_original_layers` only ever being set the very first time that thresholding changes.  That is a problem because `self._viewer.get_layers()` could be changed in between thresholding changes, leaving `_original_layers` with stale data.

hughes036: ex. 
- A directory has 100 predictions, and nothing is in the viewer
- 5 are dragged into the viewer
- thresholding changes
- another 5 are dragged in
- thresholding changes again

src/allencell_ml_segmenter/thresholding/thresholding_view.py
```diff
@@ -0,0 +1,343 @@
...
+            QSizePolicy.Policy.Preferred, QSizePolicy.Policy.Maximum
+        )
+
+        # title
+        self._title: QLabel = QLabel("THRESHOLD", self)
+        self._title.setObjectName("title")
+        layout.addWidget(self._title, alignment=Qt.AlignmentFlag.AlignHCenter)
+
+        # selecting input image
+        self._file_input_widget: FileInputWidget = FileInputWidget(
```
hughes036, line 88: Why is this part of the thresholding view?  A model is already selected in the experiment model, so we should limit the workflow to only thresholding that.  

This needs to use the guardrails that we have put in place (user experiments) or else there are too many edge cases to handle

src/allencell_ml_segmenter/services/prediction_service.py
```diff
@@ -139,7 +139,9 @@ def build_overrides(self, checkpoint: Path) -> dict[str, Any]:
         # Need these overrides to load in csv's
         overrides["data.columns"] = ["raw", "split"]
         overrides["data.split_column"] = "split"
-        overrides["ckpt_path"] = str(checkpoint)
+        overrides["checkpoint.ckpt_path"] = str(checkpoint)
+        overrides["checkpoint.strict"] = False
+        overrides["checkpoint.weights_only"] = True
```
hughes036, line 144: FYI iterative training changes, not related to pr

src/allencell_ml_segmenter/thresholding/thresholding_service.py
hughes036: Discussed a design for handling this (in person):

thresholdingModel.originalLayers

A dict of filePath (because layer name can be editted) -> probability mapping, built when prediction completes, and updated as thresholding changes

src/allencell_ml_segmenter/thresholding/thresholding_service.py
hughes036: discussed in person

src/allencell_ml_segmenter/_tests/thresholding/test_thresholding_service.py
```diff
@@ -0,0 +1,95 @@
...
+
+    # check if existing thresholds get updated
+    thresholding_model.set_thresholding_value(100)
+    assert len(viewer.get_layers()) == 1
+    seg_data = viewer.segmentation_inserted["[threshold] test_layer"]
+    assert np.array_equal(seg_data, (test_image > 100).astype(int))
+
+
+def test_on_threshold_changed_non_prediction(test_image):
+    thresholding_model: ThresholdingModel = ThresholdingModel()
```
hughes036, line 59: Im getting lost in this test - can you add a docstring explaining the scenario being simulated?

src/allencell_ml_segmenter/_tests/services/test_prediction_service.py
```diff
@@ -136,7 +136,7 @@ def test_build_overrides() -> None:
     assert overrides["train"] == False
     assert overrides["mode"] == "predict"
     assert overrides["task_name"] == "predict_task_from_app"
-    assert overrides["ckpt_path"] == str(
+    assert overrides["checkpoint.ckpt_path"] == str(
```
hughes036, line 139: another incremental training change?

src/allencell_ml_segmenter/core/file_input_widget.py
```diff
@@ -51,9 +51,10 @@ def __init__(
         model: FileInputModel,
         viewer: IViewer,
         service: ModelFileService,
+        include_channel_selection: bool = True,
     ):
         super().__init__()
-
+        self._include_channel_selection: bool = include_channel_selection
```
hughes036, line 57: Are these changes generalizing the fileInputWidget to not always deal with channels (for this use case)?

src/allencell_ml_segmenter/core/file_input_widget.py
```diff
@@ -214,19 +214,24 @@ def _from_directory_slot(self) -> None:
         self._model.set_input_mode(InputMode.FROM_PATH)
 
     def _update_layer_list(self, event: Optional[NapariEvent] = None) -> None:
+        existing_selection = [
+            self._viewer.get_layers_nonthreshold()[i].name
```
hughes036, line 218: How did the viewer become a member of the fileInputWidget?  

I see details about thresholding leaking into what is (i think) supposed to be a simple, generalized widget.

src/allencell_ml_segmenter/core/file_input_widget.py
```diff
@@ -261,10 +266,11 @@ def _process_checked_signal(self, row: int, state: Qt.CheckState) -> None:
                 )
 
     def _reset_channel_combobox(self) -> None:
-        self._channel_select_dropdown.clear()
-        self._channel_select_dropdown.setPlaceholderText("")
-        self._channel_select_dropdown.setCurrentIndex(-1)
-        self._channel_select_dropdown.setEnabled(False)
+        if self._include_channel_selection:
```
hughes036, line 269: All the detail about channel selection also seem out of place here.

src/allencell_ml_segmenter/main/experiments_model.py
```diff
@@ -76,21 +76,30 @@ def get_model_checkpoints_path(
...
+            return user_exp_path / exp_name
+        return None
+
+    def get_csv_path(self) -> Optional[Path]:
+        exp_path: Optional[Path] = self._get_exp_path()
+        if exp_path is not None:
+            return exp_path / "data"
+        return None
+
+    def get_metrics_csv_path(self) -> Optional[Path]:
```
hughes036, line 92: Iterative training changes?

src/allencell_ml_segmenter/core/file_input_model.py
```diff
@@ -79,3 +79,16 @@ def set_max_channels(self, max: Optional[int]) -> None:
 
     def get_selected_paths(self) -> Optional[list[Path]]:
         return self._selected_paths
+
+    def get_input_files_as_list(self) -> List[Path]:
+        input_image_path: Optional[Path] = self.get_input_image_path()
```
hughes036, line 84: tested?

src/allencell_ml_segmenter/_tests/fakes/fake_viewer.py
```diff
@@ -103,3 +104,22 @@ def subscribe_layers_change_event(
...
             fn(FakeNapariEvent())
+
+    def get_seg_layers(self, layer_list: list[Layer]) -> list[Layer]:
+        return [
+            layer
+            for layer in self._image_layers.values()
+            if layer.name.startswith("[seg]")
+        ]
+
+    def insert_segmentation(
```
hughes036: which of these are from the interface vs for testing only?  Can you comment them?

src/allencell_ml_segmenter/main/i_experiments_model.py
```diff
@@ -55,7 +55,7 @@ def get_model_checkpoints_path(
         pass
 
     @abstractmethod
-    def get_metrics_csv_path(self) -> Path:
+    def get_metrics_csv_path(self) -> Optional[Path]:
```
hughes036, line 58: linting?

src/allencell_ml_segmenter/main/main_model.py
```diff
@@ -85,3 +86,9 @@ def set_selected_channels(
 
     def training_complete(self) -> None:
         self.dispatch(Event.PROCESS_TRAINING_COMPLETE)
+
+    def set_predictions_in_viewer(self, predictions_in_viewer: bool) -> None:
+        self._predictions_in_viewer = predictions_in_viewer
```
hughes036, line 96: Seems more like a concern of `prediction_model`?

src/allencell_ml_segmenter/main/main_widget.py
```diff
@@ -84,13 +90,16 @@ def __init__(
         self._training_model: TrainingModel = TrainingModel(
             main_model=self._model, experiments_model=self._experiments_model
         )
-        self._file_input_model: FileInputModel = FileInputModel()
+        self._prediction_file_input_model: FileInputModel = FileInputModel()
```
hughes036, line 93: nit: having contextual fileInputModels way out here does seem weird now (i think @saeliddp was concerned about this)

hughes036 (commented):
I think that thresholding is a harder problem than I realized, and that the UX we are currently trying to support both makes for tough edge cases, and also diverges from the 'experiments home' centered, linear workflow that we have going.  We discussed a number of changes is person, and i made a few more notes here.

src/allencell_ml_segmenter/main/i_viewer.py
```diff
@@ -72,3 +72,17 @@ def subscribe_layers_change_event(
         self, function: Callable[[NapariEvent], None]
     ) -> None:
         pass
+
+    @abstractmethod
+    def get_seg_layers(self, layer_list: list[Layer]) -> list[Layer]:
```
saeliddp: The parameters for this and the function below are not self-explanatory to me--a comment would be useful

pyproject.toml
yrkim98: all my testing was done on a version of 0.4.4 that had one edit in the way I am passing images to the auto threshold function. This will be released as 0.4.5, so I will update this version pin to 0.4.5 before merging once that PR is merged in cyto-dl and published on pypi.

src/allencell_ml_segmenter/_tests/fakes/fake_viewer.py
yrkim98: these are both from the interface and mimic the actual functions. We currently dont have any test-only functions here

src/allencell_ml_segmenter/_tests/services/test_prediction_service.py
yrkim98: this was changed going from 0.1.6 -> latest version of cyto-dl

src/allencell_ml_segmenter/_tests/thresholding/test_thresholding_service.py
yrkim98: https://github.com/AllenCell/allencell-segmenter-ml/pull/573/commits/86edfa7bed32d3cf7d5888461cc20aea5a74c327

this is to test that only the relevant layers produce a threshold (not raw images, only probability mappings)

src/allencell_ml_segmenter/core/file_input_model.py
yrkim98: added tests for each case

src/allencell_ml_segmenter/core/file_input_widget.py
yrkim98: correct

src/allencell_ml_segmenter/core/file_input_widget.py
yrkim98: the viewer has always been a member of the fileinputwidget.
When selecting files from napari layers, we sync the viewer layers with the available selections in the fileinputwidget.

these changes make it so that

- only non-threshold layers show up as selectable in the fileinputwidget
- previous selections are kept when images are added/removed from the viewer, which is needed so that when we update with new threshold layers, we dont delete the user's previously selected layers. 
- The above bullet point also addresses thao's comments on previous versions that we should keep layer selections even if images are added to the viewer.

src/allencell_ml_segmenter/core/file_input_widget.py
yrkim98: file input widget is also responsible for selecting the channel of the input images, so thats why this exists here.

an alternative is to make two separate widgets, eg. `FileInputWidget` and `FileInputWidgetWithChannel` and have the latter extend the base class. Thoughts?

src/allencell_ml_segmenter/main/experiments_model.py
yrkim98: These are downstream changes from a bug that fixes ExperimentsModel (Napari viewer would get spammed with warnings when we opened it and selected options)

src/allencell_ml_segmenter/main/i_experiments_model.py
yrkim98: mypy caught this

src/allencell_ml_segmenter/main/main_model.py
yrkim98: `PredictionModel` is only for prediction-relevant state. This is moreso tracking the overall app state.
Added a comment to clarify this- this will track if there are prediction results displayed in the viewer (so triggers when prediction finishes, if predicted from images in viewer, and those images were successfully produced and are now displayed in napari)

src/allencell_ml_segmenter/thresholding/thresholding_view.py
yrkim98: Discussed last week

I agree, this allows the user to do a lot of things that we dont want to support in our workflows, but this was the UI/UX provided to me by thao.

We should go over this UX once we get this out, and iterate with her to fix this in the future. For now, due to needing to release this, I will choose to leave as is and we can revisit this later on
https://github.com/AllenCell/allencell-segmenter-ml/issues/576

src/allencell_ml_segmenter/thresholding/thresholding_view.py
yrkim98: the model inits with default values.

src/allencell_ml_segmenter/thresholding/thresholding_view.py
yrkim98: we should have full coverage on this:
https://github.com/AllenCell/allencell-segmenter-ml/blob/86edfa7bed32d3cf7d5888461cc20aea5a74c327/src/allencell_ml_segmenter/_tests/thresholding/test_thresholding_view.py#L178-L252

src/allencell_ml_segmenter/thresholding/thresholding_view.py
yrkim98: discussed: this is when thresholding is applied via the button at the bottom

src/allencell_ml_segmenter/training/image_selection_widget.py
yrkim98: mypy caught this

src/allencell_ml_segmenter/_tests/core/test_file_input_model.py
```diff
@@ -98,3 +99,120 @@ def test_set_max_channels_dispatch() -> None:
...
+    Test to see if all paths from a directory are returned as a list
+    """
+    # ARRANGE
+    file_input_model: FileInputModel = FileInputModel()
+    dummy_subscriber: FakeSubscriber = FakeSubscriber()
+    file_input_model.subscribe(
+        Event.ACTION_FILEINPUT_MAX_CHANNELS_SET,
+        dummy_subscriber,
+        dummy_subscriber.handle,
+    )
```
saeliddp: Not sure I understand why we are using a fake subscriber for the max channels set event in this test

saeliddp: I would expect the test to work just as well if you deleted lines 109-114?

saeliddp: This comment applies to all of the new tests in this file--am I missing something?

src/allencell_ml_segmenter/main/i_viewer.py
```diff
@@ -72,3 +72,31 @@ def subscribe_layers_change_event(
...
```
saeliddp: Thanks for adding comments! Still not sure on one point: Why does this need to take `layer_list` as a param? Shouldn't the viewer already have access to the layer list and apply a filter on that list to get layers that start with `[seg]`?

src/allencell_ml_segmenter/main/i_viewer.py
```diff
@@ -72,3 +72,31 @@ def subscribe_layers_change_event(
...
+        """
+        Get only segmentation layers (which should be probability mappings) from the viewer.
+        These are the layers that start with [seg].
+        """
+        pass
+
+    @abstractmethod
+    def insert_threshold(
+        self, layer_name: str, img: np.ndarray, seg_layers: bool = False
```
saeliddp, line 86: Description for `seg_layers` param?

src/allencell_ml_segmenter/main/viewer.py
```diff
@@ -105,3 +116,47 @@ def _get_layer_by_name(self, name: str) -> Optional[Layer]:
             if l.name == name:
                 return l
         return None
+
+    def get_seg_layers(self, layer_list: list[Layer]) -> list[Layer]:
```
saeliddp: Appears that `layer_list` parameter is unused in the implementation -> prob fine to get rid of it in the `i_viewer` and implementations

src/allencell_ml_segmenter/main/viewer.py
```diff
@@ -105,3 +116,47 @@ def _get_layer_by_name(self, name: str) -> Optional[Layer]:
...
+            if seg_layers:
+                seg_layer_og = self._get_layer_by_name(layer_name)
+
+            # figure out where to insert the new thresholded layer (on top of the original segmentation image)
+            layerlist_pos = layerlist.index(layer_name)
+            labels_created = Labels(image, name=f"[threshold] {layer_name}")
+            layerlist.insert(layerlist_pos + 1, labels_created)
+
+            # remove the original segmentation layer if it exists
+            if seg_layer_og:
```
saeliddp, line 164: so, `seg_layers` is ultimately a flag for whether to remove corresponding segmentation layers after thresholding? Maybe rename to `remove_seg_layers`?

src/allencell_ml_segmenter/main/viewer.py
```diff
@@ -105,3 +116,47 @@ def _get_layer_by_name(self, name: str) -> Optional[Layer]:
...
+            # check if the original segementation layer is currently in the viewer, if so, remove later after
+            # thresholding is applied
+            seg_layer_og: Optional[Layer] = None
+            if seg_layers:
+                seg_layer_og = self._get_layer_by_name(layer_name)
+
+            # figure out where to insert the new thresholded layer (on top of the original segmentation image)
+            layerlist_pos = layerlist.index(layer_name)
+            labels_created = Labels(image, name=f"[threshold] {layer_name}")
+            layerlist.insert(layerlist_pos + 1, labels_created)
```
saeliddp, line 161: Kinda frustrating that napari's viewer doesn't support inserting as a method on the viewer

src/allencell_ml_segmenter/main/viewer.py
```diff
@@ -105,3 +116,47 @@ def _get_layer_by_name(self, name: str) -> Optional[Layer]:
...
+            layerlist = self.viewer.layers
+
+            # check if the original segementation layer is currently in the viewer, if so, remove later after
+            # thresholding is applied
+            seg_layer_og: Optional[Layer] = None
+            if seg_layers:
+                seg_layer_og = self._get_layer_by_name(layer_name)
+
+            # figure out where to insert the new thresholded layer (on top of the original segmentation image)
+            layerlist_pos = layerlist.index(layer_name)
```
saeliddp, line 159: Maybe handle the case where there is no layer with `layer_name`?

src/allencell_ml_segmenter/main/segmenter_layer.py
```diff
@@ -14,9 +14,16 @@ class ShapesLayer(SegmenterLayer):
...
+class Source(SegmenterLayer):
+    path: Optional[Path] = None
+
+
 @dataclass
 class ImageLayer(SegmenterLayer):
-    path: Optional[Path]
+    path: Optional[Path] = None
+    data: Optional[np.ndarray] = None
+    source: Source = Source(name="sourcename")
```
saeliddp, line 26: The meaning of 'source' is not clear to me from this file alone. Comments would be helpful. And what is the purpose of the kwarg 'name'?

src/allencell_ml_segmenter/_tests/fakes/fake_viewer.py
```diff
@@ -24,9 +24,10 @@ def __init__(self, viewer: Optional[napari.Viewer] = None):
         self._shapes_layers: Dict[str, ShapesLayer] = {}
         self._labels_layers: Dict[str, LabelsLayer] = {}
         self._on_layers_change_fns: List[Callable] = []
+        self.segmentation_inserted: Dict[str, np.ndarray] = {}
```
saeliddp: This name is a bit misleading--this is only used for thresholds, not the initial segmentations, right? Maybe better to call it `_thresholds_inserted` or something

src/allencell_ml_segmenter/thresholding/thresholding_model.py
```diff
@@ -0,0 +1,77 @@
...
+        Get the thresholding value.
+        """
+        return self._thresholding_value_selected
+
+    def set_autothresholding_enabled(self, enable: bool) -> None:
+        """
+        Set autothresholding enabled.
+        """
+        self._is_autothresholding_enabled = enable
+        self.dispatch(Event.ACTION_THRESHOLDING_AUTOTHRESHOLDING_SELECTED)
```
saeliddp: Should this only get fired when `enable` is true?

src/allencell_ml_segmenter/_tests/thresholding/test_thresholding_view.py
```diff
@@ -0,0 +1,270 @@
...
+
+@pytest.fixture
+def thresholding_model() -> ThresholdingModel:
+    model = ThresholdingModel()
+    model.set_thresholding_value(128)
+    return model
+
+
+@pytest.fixture
+# tmp_path is a builtin pytest fixture for a faked path
```
saeliddp, line 38: Cool!

src/allencell_ml_segmenter/core/file_input_widget.py
saeliddp: ^^ that idea sounds good to me, if it's not too tricky to implement widget inheritance

src/allencell_ml_segmenter/core/file_input_widget.py
saeliddp: We don't need to deal with channels because segmentations are already single channel, right? Is there any harm in just having the channel selection option there but with only one option?

yrkim98: It would have to be one more action, they would have to select channel 0, which Thao wants to avoid

src/allencell_ml_segmenter/main/i_viewer.py
yrkim98: removed `layer_list` as param, carried over from another method i had.

src/allencell_ml_segmenter/main/i_viewer.py
yrkim98: Added!

src/allencell_ml_segmenter/main/viewer.py
yrkim98: removed

src/allencell_ml_segmenter/main/viewer.py
yrkim98: renamed to `remove_seg_layers`

src/allencell_ml_segmenter/main/viewer.py
yrkim98: yeah.. I dont think they expected people to interact with the layerlist much and instead perform operation on the images in the viewer. We do a lot of inserting/removing. Maybe that means we might need to rethink our design 🤷‍♂️

src/allencell_ml_segmenter/main/viewer.py
yrkim98: So if `insert_threshold()` is called we can assume the `layer_name` is valid. I could throw an exception here if it doesnt exist but `layerlist.index()` will do that for us instead. We shouldnt expect invalid layer names in this method so i think as-is fine for now

src/allencell_ml_segmenter/main/segmenter_layer.py
yrkim98: these just mimic the fields in the actual napari `ImageLayer` and `Source` objects, and were added for our fakes. the kwarg name exists in the actual `Source` object which our fake uses,

src/allencell_ml_segmenter/_tests/fakes/fake_viewer.py
yrkim98: totally agree, changed

src/allencell_ml_segmenter/thresholding/thresholding_model.py
yrkim98: yes! fixed.

src/allencell_ml_segmenter/_tests/thresholding/test_thresholding_view.py
yrkim98: 🎉

src/allencell_ml_segmenter/training/view.py
```diff
@@ -270,10 +270,18 @@ def train_btn_handler(self) -> None:
         if self._patch_size_ok():
             self.set_patch_size()
             num_epochs: Optional[int] = self._training_model.get_num_epochs()
+            metrics_csv_path: Optional[Path] = (
```
saeliddp, line 273: Why are there training changes in this PR?

src/allencell_ml_segmenter/thresholding/thresholding_service.py
```diff
@@ -0,0 +1,142 @@
...
+        # determine thresholding function to use
+        if self._thresholding_model.is_autothresholding_enabled():
+            thresh_function: Callable = AutoThreshold(
+                self._thresholding_model.get_autothresholding_method()
+            )
+        else:
+            thresh_function = self._threshold_image
+        for layer in layers_to_threshold:
+            # Creating helper functions for mypy strict typing
+            def thresholding_task() -> np.ndarray:
```
saeliddp, line 100: Does this get around the issue w/ lambda functions?

src/allencell_ml_segmenter/thresholding/thresholding_service.py
```diff
@@ -0,0 +1,142 @@
...
+        if self._thresholding_model.is_autothresholding_enabled():
+            thresh_function: Callable = AutoThreshold(
+                self._thresholding_model.get_autothresholding_method()
+            )
+        else:
+            thresh_function = self._threshold_image
+        for path in images_to_threshold:
+            image = BioImage(path)
+            try:
+                self._save_thresh_image(thresh_function(image.data), path.name)
```
saeliddp, line 133: Is the reason we apply thresh function again is because it's possible that they've selected a folder to threshold and the thresholded images aren't already in the viewer?

src/allencell_ml_segmenter/training/view.py
yrkim98: mypy caught this

yrkim98: (as a result of ExperimentModel changes)

src/allencell_ml_segmenter/thresholding/thresholding_service.py
yrkim98: Its effectively the same as the lambda function I had before, but adds typing to satisfy mypy. seems to be the accepted solution

src/allencell_ml_segmenter/thresholding/thresholding_service.py
yrkim98: exactly

saeliddp (approved):

src/allencell_ml_segmenter/main/segmenter_layer.py
saeliddp: Where do we use the 'source' attribute in the code?

src/allencell_ml_segmenter/thresholding/thresholding_service.py
yrkim98: changed to OrderedDict of layer_name -> probability mapping

src/allencell_ml_segmenter/thresholding/thresholding_service.py
yrkim98: This was simplified with the above change

yrkim98: https://github.com/AllenCell/allencell-segmenter-ml/pull/573/commits/d17504570cabbd8f5f99d277a689a64f7880c94b

src/allencell_ml_segmenter/thresholding/thresholding_service.py
yrkim98: This function only deals with images from the viewer. simplified here: https://github.com/AllenCell/allencell-segmenter-ml/pull/573/commits/d17504570cabbd8f5f99d277a689a64f7880c94b

src/allencell_ml_segmenter/thresholding/thresholding_service.py
yrkim98: it will overwrite their existing image.

src/allencell_ml_segmenter/thresholding/thresholding_view.py
yrkim98: comments added https://github.com/AllenCell/allencell-segmenter-ml/pull/573/commits/bcb0d8b5ce2ed3f133ceb9da41d501b02719c74a

src/allencell_ml_segmenter/thresholding/thresholding_service.py
yrkim98: https://github.com/AllenCell/allencell-segmenter-ml/pull/573/commits/d17504570cabbd8f5f99d277a689a64f7880c94b

src/allencell_ml_segmenter/core/file_input_widget.py
yrkim98: No issue, but they would have to perform one more step (selecting channels) which i believe thao wants to avoid

src/allencell_ml_segmenter/core/file_input_widget.py
yrkim98: I feel like thresholding will undergo some UX changes / rethinking post release. Im gonna postpone for now and create a ticket https://github.com/AllenCell/allencell-segmenter-ml/issues/577

src/allencell_ml_segmenter/core/file_input_model.py
yrkim98: https://github.com/AllenCell/allencell-segmenter-ml/pull/573/commits/617dbb1297172ce69a68aae2e1fd4e4b6f59d46a

src/allencell_ml_segmenter/main/main_model.py
yrkim98: comments added for clarification https://github.com/AllenCell/allencell-segmenter-ml/pull/573/commits/bc95b0ff36e7eee322ea5b8fa220526d96366437

src/allencell_ml_segmenter/main/main_widget.py
yrkim98: This would be a significant refactor, so deciding to not work on for now.

yrkim98: https://github.com/AllenCell/allencell-segmenter-ml/issues/578

src/allencell_ml_segmenter/_tests/core/test_file_input_model.py
yrkim98: you're right! removed https://github.com/AllenCell/allencell-segmenter-ml/pull/573/commits/6d828e09128327996649a6e4af6c4e8819093b2e

src/allencell_ml_segmenter/main/segmenter_layer.py
yrkim98: `_update_layer_list()` in `FileInputWidget` uses this. the kwarg name is a required argument to Source.

src/allencell_ml_segmenter/core/file_input_widget.py
hughes036: FWIW i think ideally the interaction between fileInputWidgets and the Viewer happens via events.

PR #575, opened 2024-12-04
Staging

yrkim98:
## Purpose
Since we released 1.0.0rc0 from staging with hotfixes cherry-picked from main, these two should now both have all the changes for iterative training. 
However, i checked the diff and it looks like I previously made a PR directly onto staging so main is missing a few of these needed changes. Main also does not contain the latest version bump.

This makes sure that these two are synced at the moment with all changes that have passed Thao's QC. From this point onwards, unless a hotfix is needed for iterative training, we will do our releases from `main`.

If hotfixes for iterative training are needed- I will keep the staging branch so we can continue to update as needed. This PR will merge the code that connects the plugin to the `STG_BUCKET`, but I will change this as we release 1.0.0 to `PROD_BUCKET`.

## Changes
Syncs all iterative training hotfixes that thao has GL'ed into main, updates version pin.

## Testing
Passes all previous tests

## How to review
Make sure that the diff looks okay. Since most of the iterative training hotfixes are already on main, this should only contain the following:

- version pin change
- staging bucket change
- creating a folder for the prediction csv.


pyproject.toml
```diff
@@ -105,8 +105,8 @@ where = ["src"]
 
 # https://pypi.org/project/bumpver
 [tool.bumpver]
-current_version = "0.1.16"
-version_pattern = "MAJOR.MINOR.PATCH[PYTAGNUM]"
+current_version = "1.0.0rc0"
+version_pattern = "MAJOR.MINOR.PATCH[.PYTAGNUM]"
```
saeliddp: I don't think we want the `.` added back in; iirc, PyPI only recognizes RC versions without the `.`

yrkim98: https://github.com/AllenCell/allencell-segmenter-ml/pull/575/commits/15bdec324ecf4cdddfee2081e80c3d83e5635cd6

good catch, thanks

saeliddp (approved):

PR #581, opened 2025-01-17
Refactor/move original layer data to layer metadata

yrkim98:
## Purpose
- This is a refactor of the thresholding code- enforces that users only threshold images that are generated in the plugin, rather than allowing thresholding of any image dragged into viewer.
- how it works: when segmentations are generated in the prediction part of the plugin, they are saved with the napari viewer metadata `prob_map` which is the original probability mapping before any thresholding is applied.
- Thresholding looks for all layers with a `prob_map` key in its metadata- which indicates it is a segmentation generated from our plugin
- thresholding is performed with `prob_map` as the input.
- This will need some UI changes to work completely, which will come in a follow up PR (i will stop using the FileInputWidget, which allows the user to select any image dropped in napari, and instead only allow selection of images generated in prediction)

## Changes
- Main change is removing the `original_layers_in_viewer` field from the thresholding model. We instead track the original probability mappings through the layer metadata itself.
- Includes changes in other parts of the plugin that used to depend on `original_layers_in_viewer` field. Uses the napari viewer metadata instead now.

## Testing
Tested on windows

## How to review
- start with `src/allencell_ml_segmenter/thresholding/thresholding_model.py` which contains the changes to get rid of the `original_layers_in_viewer` field.
- review `src/allencell_ml_segmenter/thresholding/thresholding_service.py` which contains most of the changes that are affected by changing the way we store prediction outputs as napari viewer metadata rather than as the `original_layers_in_viewer` field.
- review other files that have changed as a result of this model change

### DIFF
--- a/src/allencell_ml_segmenter/_tests/fakes/fake_viewer.py
+++ b/src/allencell_ml_segmenter/_tests/fakes/fake_viewer.py
@@ -13,7 +13,7 @@ from allencell_ml_segmenter.main.segmenter_layer import (
 from napari.utils.events import Event as NapariEvent
 import napari
 from typing import List, Dict, Callable, Optional
-from napari.layers import Layer
+from napari.layers import Layer, Labels
 
 
 class FakeNapariEvent:
@@ -137,3 +137,10 @@ class FakeViewer(IViewer):
             return Path(layer.metadata["source_path"])
 
         return None
+
+    # TODO: add these fakes for testing methods that rely on this
+    def add_segmentation_labels(self, masked_data: np.ndarray, name: str, prob_map: np.ndarray) -> None:
+        pass
+
+    def get_all_segmentation_labels(self) -> list[Labels]:
+        pass
--- a/src/allencell_ml_segmenter/_tests/fakes/fake_viewer.py
+++ b/src/allencell_ml_segmenter/_tests/fakes/fake_viewer.py
@@ -139,8 +139,15 @@ class FakeViewer(IViewer):
         return None
 
     # TODO: add these fakes for testing methods that rely on this
-    def add_segmentation_labels(self, masked_data: np.ndarray, name: str, prob_map: np.ndarray) -> None:
+    def add_segmentation_labels(
+        self, masked_data: np.ndarray, name: str, prob_map: np.ndarray
+    ) -> None:
         pass
 
     def get_all_segmentation_labels(self) -> list[Labels]:
-        pass
+        return [
+            layer
+            for layer in self.get_all_images()
+            if getattr(layer, "metadata", None)
+            and "prob_map" in layer.metadata
+        ]
--- a/src/allencell_ml_segmenter/_tests/fakes/fake_viewer.py
+++ b/src/allencell_ml_segmenter/_tests/fakes/fake_viewer.py
@@ -138,12 +138,6 @@ class FakeViewer(IViewer):
 
         return None
 
-    # TODO: add these fakes for testing methods that rely on this
-    def add_segmentation_labels(
-        self, masked_data: np.ndarray, name: str, prob_map: np.ndarray
-    ) -> None:
-        pass
-
     def get_all_segmentation_labels(self) -> list[Labels]:
         return [
             layer
--- a/src/allencell_ml_segmenter/_tests/thresholding/test_thresholding_service.py
+++ b/src/allencell_ml_segmenter/_tests/thresholding/test_thresholding_service.py
@@ -68,11 +68,20 @@ def test_on_threshold_changed_non_prediction(test_image):
         viewer,
         task_executor=SynchroTaskExecutor.global_instance(),
     )
-    # Only the [seg] layers below should produce a threshold layer
+
+    # Only the [seg] layers below should produce a threshold layer since they have prob map metadata
     viewer.add_image(test_image, name="[raw] test_layer 1")
-    viewer.add_image(test_image, name="[seg] test_layer 1")
+    viewer.add_image(
+        test_image,
+        name="[seg] test_layer 1",
+        metadata={"prob_map": test_image},
+    )
     viewer.add_image(test_image, name="[raw] test_layer 2")
-    viewer.add_image(test_image, name="[seg] test_layer 2")
+    viewer.add_image(
+        test_image,
+        name="[seg] test_layer 2",
+        metadata={"prob_map": test_image},
+    )
     viewer.add_image(test_image, name="donotthreshold")
 
     # ACT set a threshold to trigger
--- a/src/allencell_ml_segmenter/main/viewer.py
+++ b/src/allencell_ml_segmenter/main/viewer.py
@@ -192,3 +192,22 @@ class Viewer(IViewer):
         if layer.metadata is not None and "source_path" in layer.metadata:
             return Path(layer.metadata["source_path"])
         return None
+
+    def add_segmentation_labels(self, masked_data: np.ndarray, name: str, prob_map: np.ndarray) -> None:
+        """
+        Add a segmentation labels layer and save probability mapping in the layer metadata.
+        """
+        self.add_labels(
+            masked_data,
+            name=name,
+            metadata={
+                "prob_map": prob_map
+            },
+        )
+
+    def get_all_segmentation_labels(self) -> list[Labels]:
+        """
+        Get all segmentation labels layers that currently exist in the viewer.
+        """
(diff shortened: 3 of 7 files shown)


src/allencell_ml_segmenter/_tests/thresholding/test_thresholding_service.py
```diff
@@ -68,11 +68,20 @@ def test_on_threshold_changed_non_prediction(test_image):
...
+        name="[seg] test_layer 1",
+        metadata={"prob_map": test_image},
+    )
     viewer.add_image(test_image, name="[raw] test_layer 2")
-    viewer.add_image(test_image, name="[seg] test_layer 2")
+    viewer.add_image(
+        test_image,
+        name="[seg] test_layer 2",
+        metadata={"prob_map": test_image},
+    )
```
yrkim98, line 84: this changes the test to mimic how we add these segmentations in the prediction portion of our plugin

src/allencell_ml_segmenter/_tests/fakes/fake_viewer.py
```diff
@@ -137,3 +137,11 @@ def get_source_path(self, layer: Layer) -> Optional[Path]:
...
 
         return None
+
+    def get_all_segmentation_labels(self) -> list[Labels]:
+        return [
+            layer
+            for layer in self.get_all_images()
+            if getattr(layer, "metadata", None)
+            and "prob_map" in layer.metadata
+        ]
```
yrkim98, line 147: the `getattr(layer, "metadata", None)` is to satisfy type checking- making sure that the metadata field exists in this faked layer.

src/allencell_ml_segmenter/thresholding/thresholding_service.py
```diff
@@ -99,18 +80,29 @@ def _on_threshold_changed(self, _: Event) -> None:
...
             def thresholding_task() -> np.ndarray:
-                return thresh_function(image)
+                # INVARIANT: a segmentation layer must have prob_map in its metadata if it came from our plugin
+                # so we are only supporting thresholding images that are from the plugin itself.
+                if (
+                    not isinstance(layer.metadata, dict)
+                    or "prob_map" not in layer.metadata
+                ):
+                    raise ValueError(
+                        "Layer metadata must be a dictionary containing the 'prob_map' key in order to threshold."
```
yrkim98, line 93: this is again to satisfy mypy. It should be impossible to reach this line of code since the `self._viewer.get_all_segmentation_labels()` function above only grabs layers with "prob_map" in its metadata, and the napari layer will always have a metadata field.

src/allencell_ml_segmenter/thresholding/thresholding_service.py
```diff
@@ -99,18 +80,29 @@ def _on_threshold_changed(self, _: Event) -> None:
...
             def on_return(
-                thresholded_image: np.ndarray,
-                layer_name_instance: str = layer_name,
+                threshold_output: np.ndarray,
             ) -> None:
                 self._viewer.insert_threshold(
-                    layer_name_instance,
-                    thresholded_image,
+                    layer_instance.name,
+                    threshold_output,
```
yrkim98, line 105: this will change slightly with my following UI changes. Since we keep track of the prob map in the layer metadata, rather than replacing the layer with a new layer we want to replace the data of the layer instead.

src/allencell_ml_segmenter/main/viewer.py
```diff
@@ -192,3 +192,13 @@ def get_source_path(self, layer: Layer) -> Optional[Path]:
         if layer.metadata is not None and "source_path" in layer.metadata:
             return Path(layer.metadata["source_path"])
         return None
+
+    def get_all_segmentation_labels(self) -> list[LabelsLayer]:
```
saeliddp, line 196: I think originally the idea with `Viewer` was to have it wrap napari viewer functionality so that we could fake it for testing. The concept of a segmentation label is not a napari-specific thing, but since we are manipulating the layer's metadata, maybe it makes sense for that to be here? Idk just something to consider

src/allencell_ml_segmenter/thresholding/thresholding_service.py
saeliddp: Makes me wonder if mypy is a net positive or not for python projects...

saeliddp (approved):
Seems like a nice simplification compared to the base PR

src/allencell_ml_segmenter/main/viewer.py
yrkim98: I realized this when working with this function, and fixed it in my follow up PR(which will be out shortly!)

src/allencell_ml_segmenter/thresholding/thresholding_service.py
yrkim98: its recommended... but in spots like this i agree its just overkill. 

I also cant think of a way to rewrite this so we wouldn't need this check, if you can lmk

src/allencell_ml_segmenter/main/viewer.py
hughes036: I'd agree that this makes more sense in our `thresholding_service`, leaving our `view` as just an abstraction of napari viewer methods.

src/allencell_ml_segmenter/thresholding/thresholding_service.py
```diff
@@ -64,32 +65,12 @@ def __init__(
...
-                self._viewer.get_layers()
-            )
-
-        # Get layers to threshold.
-        # if there are segmentations displayed in the viewer, only threshold those images.
-        layers_to_threshold: OrderedDict[str, np.ndarray] = (
-            self._thresholding_model.get_layers_to_threshold(
-                self._main_model.are_predictions_in_viewer()
-            )
+        segmentation_labels: list[LabelsLayer] = (
```
hughes036, line 72: nit: Do we need to copy this to local state on threshold changes?  Couldn't users of `segmentation_labels` just call `self._viewer.get_all_segmentation_labels()`?

src/allencell_ml_segmenter/thresholding/thresholding_service.py
```diff
@@ -99,18 +80,29 @@ def _on_threshold_changed(self, _: Event) -> None:
...
-        for layer_name, image in layers_to_threshold.items():
+        for layer in segmentation_labels:
             # Creating helper functions for mypy strict typing
             def thresholding_task() -> np.ndarray:
-                return thresh_function(image)
+                # INVARIANT: a segmentation layer must have prob_map in its metadata if it came from our plugin
+                # so we are only supporting thresholding images that are from the plugin itself.
+                if (
+                    not isinstance(layer.metadata, dict)
+                    or "prob_map" not in layer.metadata
```
hughes036, line 90: isnt self._viewer.get_all_segmentation_labels already responsible for this invariant?

src/allencell_ml_segmenter/thresholding/thresholding_service.py
hughes036: yeah it seems like this promotes choosing between:
1. not DRY code
2. not failing fast

src/allencell_ml_segmenter/thresholding/thresholding_service.py
```diff
@@ -99,18 +80,29 @@ def _on_threshold_changed(self, _: Event) -> None:
...
+                return thresh_function(layer.metadata["prob_map"])
+
+            layer_instance: LabelsLayer = layer
 
             def on_return(
-                thresholded_image: np.ndarray,
-                layer_name_instance: str = layer_name,
+                threshold_output: np.ndarray,
             ) -> None:
                 self._viewer.insert_threshold(
```
hughes036, line 103: naming nit: Is this replacing the binary map?

src/allencell_ml_segmenter/thresholding/thresholding_service.py
```diff
@@ -99,18 +80,29 @@ def _on_threshold_changed(self, _: Event) -> None:
...
+                # so we are only supporting thresholding images that are from the plugin itself.
+                if (
+                    not isinstance(layer.metadata, dict)
+                    or "prob_map" not in layer.metadata
+                ):
+                    raise ValueError(
+                        "Layer metadata must be a dictionary containing the 'prob_map' key in order to threshold."
+                    )
+
+                return thresh_function(layer.metadata["prob_map"])
```
hughes036, line 96: comment what this returns (a binary map?)

hughes036 (approved):

PR #583, opened 2025-02-14
Make ThresholdingModel compose FileInputModel, Add ability to remove binary maps

yrkim98:
## Purpose
This PR resolves two things:
- Refactor:`ThresholdingModel` now inherits `FileInputModel`, so that all UI state is reflected in a single location (`ThresholdingModel`). `ThresholdingService` references state in a single location (`ThresholdingModel`), updating the viewer as the user selects/unselects images and thresholding methods.

- Bugfix: unselecting an image does not clear the binary map for that image.
- Bugfix: selecting (None) in the UI does not clear out all binary maps generated from thresholding

A short video demo of the last two points:
https://github.com/user-attachments/assets/cf1e6a9d-2b21-41bb-90fe-e494a9b54fb3

## Changes
- `ThresholdingModel` now extends `FileInputModel`, this is the single source of truth for thresholding state. before, we stored state from `ThresholdingFileInputWidget` and `ThresholdingView` in separate places. Now this model stores all thresholding state.
- Any UI updates that should trigger a new binary map to be generated or removed from the viewer fire a single event, `Event.ACTION_EXECUTE_THRESHOLDING`. Based on UI state, the `ThresholdingService` handles this event, generates binary maps, and updates the viewer as needed.
- When unselecting an image, the binary map for that image is cleared in the napari viewer.
- When selecting "None" for the thresholding methods in the viewer, all binary maps are cleared.

## Testing
Unit tests, tested on windows

## How to review
1. Start with `ThresholdingModel` and see how it composes `FileInputModel` to use with `ThresholdingFileInputWidget` as well as the `ThresholdingView`
2. See the event handler in `ThresholdingService._on_threshold_changed` which handles all threshold events and updates the viewer as needed. This now includes logic to remove binary maps from the viewer when they are unselected.
3. There is a new `Viewer` function to remove a binary map from a layer. This is done by creating a image of all zeros using `np.zeros()` and adding that as a placeholder when we do not want to display any binary map in a layer. We want to keep the layer, since it contains the original probability map in its metadata that we will need if this layer needs to be thresholded again.
4. `TestThresholdingService` contains two new tests which ensure that images are removed when they are unselected or if None is selected as the thresholding method.

### DIFF
--- a/src/allencell_ml_segmenter/_tests/thresholding/test_thresholding_service.py
+++ b/src/allencell_ml_segmenter/_tests/thresholding/test_thresholding_service.py
@@ -82,10 +82,10 @@ def test_on_threshold_changed_non_prediction(test_image):
         metadata={"prob_map": test_image},
     )
     viewer.add_image(test_image, name="donotthreshold")
-    thresholding_model.set_selected_idx([0, 1])
     thresholding_model.set_thresholding_layers(
         viewer.get_all_layers_containing_prob_map()
     )
+    thresholding_model.set_selected_idx([0, 1])
 
     # ACT set a threshold to trigger
     thresholding_model.set_thresholding_value(50)
--- a/src/allencell_ml_segmenter/_tests/thresholding/test_thresholding_service.py
+++ b/src/allencell_ml_segmenter/_tests/thresholding/test_thresholding_service.py
@@ -110,3 +110,121 @@ def test_on_threshold_changed_non_prediction(test_image):
     assert np.array_equal(seg_data, (test_image > 100).astype(int))
     # verify that raw layers do not get thresholded
     assert len(viewer.threshold_inserted) == 2
+
+
+def test_none_selected_removes_binary_maps(test_image):
+    """
+    Test that the thresholding service does not add a threshold layer for a layer that is not a probability map
+    """
+    # ARRANGE
+    thresholding_model: ThresholdingModel = ThresholdingModel()
+    viewer: FakeViewer = FakeViewer()
+    main_model: MainModel = MainModel()
+    main_model.set_predictions_in_viewer(True)
+    thresholding_service: ThresholdingService = ThresholdingService(
+        thresholding_model,
+        FakeExperimentsModel(),
+        main_model,
+        viewer,
+        task_executor=SynchroTaskExecutor.global_instance(),
+    )
+
+    # Only the [seg] layers below should produce a threshold layer since they have prob map metadata
+    viewer.add_image(test_image, name="[raw] test_layer 1")
+    viewer.add_image(
+        test_image,
+        name="[seg] test_layer 1",
+        metadata={"prob_map": test_image},
+    )
+    viewer.add_image(test_image, name="[raw] test_layer 2")
+    viewer.add_image(
+        test_image,
+        name="[seg] test_layer 2",
+        metadata={"prob_map": test_image},
+    )
+    viewer.add_image(test_image, name="donotthreshold")
+    thresholding_model.set_thresholding_layers(
+        viewer.get_all_layers_containing_prob_map()
+    )
+    thresholding_model.set_selected_idx([0, 1])
+
+    # create binary maps and add to viewer
+    thresholding_model.set_thresholding_value(50)
+    thresholding_model.set_threshold_enabled(True)
+
+    # Verify a threshold layer is added for each seg layer
+    assert "[threshold] [seg] test_layer 1" in viewer.threshold_inserted
+    seg_data = viewer.threshold_inserted["[threshold] [seg] test_layer 1"]
+    assert np.array_equal(seg_data, (test_image > 50).astype(int))
+    assert "[threshold] [seg] test_layer 2" in viewer.threshold_inserted
+    seg_data = viewer.threshold_inserted["[threshold] [seg] test_layer 2"]
+    assert np.array_equal(seg_data, (test_image > 50).astype(int))
+    # verify that raw layers do not get thresholded
+    assert len(viewer.threshold_inserted) == 2
+
+    # ACT remove all binary maps by selecting None
+    thresholding_model.disable_all()
+
+    # Assert
+    assert len(viewer.threshold_inserted) == 0
+
+
+def test_unselection_removes_binary_maps(test_image):
+    """
+    Test that the thresholding service does not add a threshold layer for a layer that is not a probability map
+    """
+    # ARRANGE
+    thresholding_model: ThresholdingModel = ThresholdingModel()
+    viewer: FakeViewer = FakeViewer()
+    main_model: MainModel = MainModel()
+    main_model.set_predictions_in_viewer(True)
+    thresholding_service: ThresholdingService = ThresholdingService(
+        thresholding_model,
+        FakeExperimentsModel(),
+        main_model,
+        viewer,
+        task_executor=SynchroTaskExecutor.global_instance(),
+    )
+
+    # Only the [seg] layers below should produce a threshold layer since they have prob map metadata
+    viewer.add_image(test_image, name="[raw] test_layer 1")
+    viewer.add_image(
+        test_image,
+        name="[seg] test_layer 1",
+        metadata={"prob_map": test_image},
+    )
+    viewer.add_image(test_image, name="[raw] test_layer 2")
+    viewer.add_image(
+        test_image,
+        name="[seg] test_layer 2",
+        metadata={"prob_map": test_image},
+    )
+    viewer.add_image(test_image, name="donotthreshold")
+    thresholding_model.set_thresholding_layers(
+        viewer.get_all_layers_containing_prob_map()
+    )
+    thresholding_model.set_selected_idx([0, 1])
+
+    # create binary maps and add to viewer
+    thresholding_model.set_thresholding_value(50)
+    thresholding_model.set_threshold_enabled(True)
+
+    # Verify a threshold layer is added for each seg layer
+    assert "[threshold] [seg] test_layer 1" in viewer.threshold_inserted
+    seg_data = viewer.threshold_inserted["[threshold] [seg] test_layer 1"]
+    assert np.array_equal(seg_data, (test_image > 50).astype(int))
+    assert "[threshold] [seg] test_layer 2" in viewer.threshold_inserted
+    seg_data = viewer.threshold_inserted["[threshold] [seg] test_layer 2"]
+    assert np.array_equal(seg_data, (test_image > 50).astype(int))
+    # verify that raw layers do not get thresholded
+    assert len(viewer.threshold_inserted) == 2
+
+    # ACT remove binary map by unselecting index 1
+    thresholding_model.set_selected_idx(
+        [0]
+    )  # remove 1, which was previously selected
+
+    # Assert
+    assert len(viewer.threshold_inserted) == 1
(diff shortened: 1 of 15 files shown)


src/allencell_ml_segmenter/_tests/thresholding/test_thresholding_service.py
```diff
@@ -110,3 +110,121 @@ def test_on_threshold_changed_non_prediction(test_image):
     assert np.array_equal(seg_data, (test_image > 100).astype(int))
     # verify that raw layers do not get thresholded
     assert len(viewer.threshold_inserted) == 2
+
+
+def test_none_selected_removes_binary_maps(test_image):
```
yrkim98, line 115: This tests if selecting None removes all binary maps from the viewer

src/allencell_ml_segmenter/_tests/thresholding/test_thresholding_service.py
```diff
@@ -110,3 +110,121 @@ def test_on_threshold_changed_non_prediction(test_image):
...
+    assert len(viewer.threshold_inserted) == 2
+
+    # ACT remove all binary maps by selecting None
+    thresholding_model.disable_all()
+
+    # Assert
+    assert len(viewer.threshold_inserted) == 0
+
+
+def test_unselection_removes_binary_maps(test_image):
```
yrkim98, line 172: This tests if unselecting an image in the ThresholdingFileInputWidget removes the binary map for that layer in the viewer

src/allencell_ml_segmenter/core/event.py
```diff
@@ -65,9 +65,9 @@ class Event(Enum):
     ACTION_CURATION_RAW_THREAD_ERROR = "curation_raw_thread_error"
 
     # Thresholding events
-    ACTION_THRESHOLDING_VALUE_CHANGED = "thresholding_value_changed"
-    ACTION_THRESHOLDING_AUTOTHRESHOLDING_SELECTED = "autothresholding_selected"
+    ACTION_EXECUTE_THRESHOLDING = "execute_thresholding"
     ACTION_SAVE_THRESHOLDING_IMAGES = "save_thresholding_images"
+    ACTION_THRESHOLDING_DISABLED = "disable_thresholding"
```
yrkim98, line 70: I use a second event that dispatches when None is selected in the UI for thresholding. This is separated to prevent the user from losing binary maps from the viewer if they unselect all thresholding method. You have to specifically select "None" to clear the viewer

src/allencell_ml_segmenter/core/file_input_widget.py
```diff
@@ -50,14 +50,14 @@ def __init__(
         self,
         model: FileInputModel,
         viewer: IViewer,
-        service: ModelFileService,
+        service: Optional[ModelFileService],
```
yrkim98, line 53: `ModelFileService` is only used the the base class, `FileInputWidget`, and not `ThresholdingInputWidget`, so these changes make this an optional parameter which will be `None` for `ThresholdingInputWidget`. The `None` check below satisfies mypy type checking

src/allencell_ml_segmenter/main/viewer.py
```diff
@@ -185,3 +188,16 @@ def get_all_layers_containing_prob_map(self) -> list[Layer]:
             for layer in self.get_layers()
             if "prob_map" in layer.metadata
         ]
+
+    def clear_binary_map_from_layer(self, layer: Layer) -> None:
```
yrkim98, line 192: viewer function to clear the binary map from a layer by replacing it with a mask containing all 0's, which will result in a transparent layer in the napari viewer.

I do this to not remove that layer completely from the viewer, which we need to keep.

src/allencell_ml_segmenter/thresholding/thresholding_model.py
```diff
@@ -72,12 +75,32 @@ def set_threshold_enabled(self, enabled: bool) -> None:
...
+        if enabled:
+            self.dispatch(Event.ACTION_EXECUTE_THRESHOLDING)
 
     def is_threshold_enabled(self) -> bool:
         """
         Get threshold specific value.
         """
         return self._is_threshold_enabled
 
+    def disable_all(self) -> None:
```
yrkim98, line 87: function called when None is selected in the UI

src/allencell_ml_segmenter/thresholding/thresholding_service.py
```diff
@@ -48,83 +44,97 @@ def __init__(
...
+            Event.ACTION_THRESHOLDING_DISABLED,
             self,
-            self._save_thresholded_images,
+            self._remove_all_prob_maps,
         )
 
     def _handle_thresholding_error(self, error: Exception) -> None:
         show_info("Thresholding failed: " + str(error))
 
     def _on_threshold_changed(self, _: Event) -> None:
```
yrkim98, line 67: handler for thresholding events

src/allencell_ml_segmenter/core/file_input_widget.py
hughes036, line 53: I think exposing a second constructor for the subclasses would be a nice way to get around making `service` Optional.

src/allencell_ml_segmenter/core/file_input_widget.py
```diff
@@ -271,7 +271,8 @@ def _process_checked_signal(self, row: int, state: Qt.CheckState) -> None:
                     len(selected_indices) == 0
                     and self._include_channel_selection
                 ):
-                    self._service.stop_channel_extraction()  # stop so combobox doesn't reset after thread is finished
+                    if self._service:
+                        self._service.stop_channel_extraction()  # stop so combobox doesn't reset after thread is finished
```
hughes036, line 275: Above idea would clean this up.

src/allencell_ml_segmenter/main/viewer.py
hughes036: Details of this are worth a docstring I think

"We need to keep the layer because it contains the segmentation's probability map.  So, clear out the binary map by setting it to all zeros.."

hughes036: Also, this public method looks like a good candidate for unit tests.

src/allencell_ml_segmenter/thresholding/thresholding_model.py
hughes036: It seems strange to enable deselecting a whole radio group, and having that behavior be different from the 'none' option.

hughes036: Im confused by the radio handling - would it be simpler if there was 1 handler for the whole group?  

- [ ] check qt docs

src/allencell_ml_segmenter/thresholding/thresholding_service.py
```diff
@@ -48,83 +44,97 @@ def __init__(
...
-            self._on_threshold_changed,
+            self._save_thresholded_images,
         )
 
         self._thresholding_model.subscribe(
-            Event.ACTION_SAVE_THRESHOLDING_IMAGES,
+            Event.ACTION_THRESHOLDING_DISABLED,
             self,
-            self._save_thresholded_images,
+            self._remove_all_prob_maps,
```
hughes036: This removes all _BINARY_ maps and leaves prob maps, right?

src/allencell_ml_segmenter/thresholding/thresholding_service.py
```diff
@@ -48,83 +44,97 @@ def __init__(
...
 
     def _handle_thresholding_error(self, error: Exception) -> None:
         show_info("Thresholding failed: " + str(error))
 
     def _on_threshold_changed(self, _: Event) -> None:
-        layers_containing_prob_map: list[Layer] = (
-            self._viewer.get_all_layers_containing_prob_map()
-        )
+        # Check to see if user has selected a thresholding method
+        if (
```
hughes036, line 69: A lot of logic would be simplified by 

- not having a 'none' option
- having a default selection (auto-thresholding seems like a good default)

src/allencell_ml_segmenter/thresholding/thresholding_service.py
hughes036, line 67: Does this handle toggling between 'on-screen images' and 'image directory' ?

src/allencell_ml_segmenter/thresholding/thresholding_service.py
```diff
@@ -48,83 +44,97 @@ def __init__(
...
-                    # INVARIANT: a segmentation layer must have prob_map in its metadata if it came from our plugin
-                    # so we are only supporting thresholding images that are from the plugin itself.
-                    if (
-                        not isinstance(layer_instance.metadata, dict)
-                        or "prob_map" not in layer_instance.metadata
-                    ):
-                        raise ValueError(
-                            "Layer metadata must be a dictionary containing the 'prob_map' key in order to threshold."
+                    # On return, display the binary map that was produced from thresholding
+                    def on_return(
```
hughes036, line 114: I think this would benfit from 1 more iteration on design: remove the 'none' option.  Id like to see how that simplifies this logic

src/allencell_ml_segmenter/thresholding/thresholding_service.py
```diff
@@ -156,3 +166,7 @@ def _threshold_image(self, image: np.ndarray) -> np.ndarray:
             self._thresholding_model.get_thresholding_value()
         )
         return (image > threshold_value).astype(int)
+
+    def _remove_all_prob_maps(self, _: Event) -> None:
```
hughes036: should be renamed `_remove_all_binary_maps`

hughes036 (approved):
This is looking pretty good.  I think lets go with it (plus bug fixes identified) for a V1.0.0 RC.  Then, we can iterate on simplifying the logic by simplifying the thresholding UX.

src/allencell_ml_segmenter/main/viewer.py
yrkim98: https://github.com/AllenCell/allencell-segmenter-ml/pull/583 docstring + tests added

src/allencell_ml_segmenter/thresholding/thresholding_model.py
yrkim98: I can confirm that a radio button cannot be unselected.
Nothing about having one handler for all radio buttons in one group. https://doc.qt.io/qt-6/qradiobutton.html

That was originally what I was going for, but we need different behavior when None is selected. I suppose the logic for that can be included in a single handler. I will create a ticket for this

src/allencell_ml_segmenter/core/file_input_widget.py
yrkim98: [a ticket was created for this](https://github.com/AllenCell/allencell-segmenter-ml/issues/586)

src/allencell_ml_segmenter/thresholding/thresholding_service.py
yrkim98: [correct](https://github.com/AllenCell/allencell-segmenter-ml/pull/583)

src/allencell_ml_segmenter/thresholding/thresholding_service.py
yrkim98: [ticket created for this](https://github.com/AllenCell/allencell-segmenter-ml/issues/587)

benjijamorris (approved):

PR #598, opened 2025-03-12
ignore large files with git lfs

yrkim98:
## Purpose
Ignores large binary files (like png images) in git diffs for PR's

[Git LFS docs](https://git-lfs.com/)

## Changes
will make it so that images dont add tons of lines in pr diffs
All changes were created by running `git lfs track`. We are tracking `.png` and `.jpeg` for now with git lfs

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+.png filter=lfs diff=lfs merge=lfs -text
+.jpeg filter=lfs diff=lfs merge=lfs -text


.gitattributes
```diff
@@ -0,0 +1,2 @@
+.png filter=lfs diff=lfs merge=lfs -text
```
hughes036, line 1: How does this work?  What is `lfs` and why are the filter and merge attributes needed?  Whats the text flag for?

hughes036: discussed in person

hughes036 (approved):

saeliddp (approved):

benjijamorris (approved):

PR #616, opened 2025-04-22
Fix help text

yrkim98:
## Purpose
Fixes "Help" dropdown text not showing on windows computers.
On macOs/Linux it seems like `QComboBox.setPlaceholderText()` works as intended, but PyQt5 users report that on windows `placeholderText` does not appear unless the combo box is editable. Our dropdown is not editable since it is a selection dropdown

## Changes
in src/allencell_ml_segmenter/training/model_selection_widget.py: Fixed the dropdown placeholder text by adding it as an item at index 0, which achieves the same behavior as `placeholderText` 

## Testing
Tested on windows

## How to review
One file changed


hughes036 (approved):
Looks the same ?  Sgtm.  

Using a combo as a menu is a strange choice that i made because i couldnt seem to find a menu that did what we needed.  We should make sure that qt doesnt have a better option (seems like it must)

hughes036:
checks are also failing here

PR #618, opened 2025-04-24
Fix failing pytest-qt on newest Ubuntu image

yrkim98:
## Purpose
When github runners updated to `ubuntu-24.04@20250420.1.0`, CI started failing for it.

Here is a rundown of the issue caused by this update:

- github runners updated to a new image `ubuntu-24.04@20250420.1.0`
- this updated the system libraries `libxcb`, `libX11` (among others, but these are the problematic ones) .
- `cyto-dl` installs `opencv-python` which bundles its own `Qt` ,`libxcb` and `libX11` libraries
- These bundled libaries are incompatible with the `ubuntu-24.04@20250420.1.0` image
- When `pytest-qt` tries to start, `Qt` fails to initialize due to this incompatibility: `Could not load the Qt platform plugin "xcb"`

I debugged this by adding a few steps in our workflow file, which has since been removed: https://github.com/AllenCell/allencell-segmenter-ml/actions/runs/14652505258/job/41121362602 

## Changes
The solution is to have our workflow file uninstall `opencv-python` for testing and installing `opencv-python-headless` which wont grab the problematic `Qt` and `libxcb`.

## Testing
Passes CI now (finally)

## How to review
Just a change to `test_lint.yml`

### DIFF
--- a/.github/workflows/test_lint.yml
+++ b/.github/workflows/test_lint.yml
@@ -55,6 +55,11 @@ jobs:
         run: |
           pip install pipdeptree
           pipdeptree --reverse --warn silence | grep -A2 opencv-python || echo "opencv-python not found"
+
+      - name: Force opencv-python-headless
+        run: |
+          pip uninstall -y opencv-python
+          pip install opencv-python-headless
       - name: Test with pytest
         run: |
           pytest -v -s --color=yes --cov=allencell_ml_segmenter --cov-report=xml
--- a/.github/workflows/test_lint.yml
+++ b/.github/workflows/test_lint.yml
@@ -49,7 +49,7 @@ jobs:
           pip install opencv-python-headless
       - name: Test with pytest
         run: |
-          pytest -v -s --color=yes --cov=allencell_ml_segmenter --cov-report=xml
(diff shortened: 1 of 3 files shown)


.github/workflows/test_lint.yml
```diff
@@ -43,6 +43,10 @@ jobs:
       - name: Type check with mypy
         run: |
           mypy src/allencell_ml_segmenter
+      - name: Force opencv-python-headless
+        run: |
+          pip uninstall -y opencv-python
+          pip install opencv-python-headless
```
saeliddp, line 51: What exactly does 'headless' mean in this context?

saeliddp (approved):
Nice!

tyler-foster (approved):
I'll approve it if it works, but could you please add a comment to the file explaining why this step is necessary? Otherwise it'll be a real head scratcher in two years after we've forgotten all about it.

.github/workflows/test_lint.yml
yrkim98: "headless" refers to no GUI support. This is why this doesnt include any GUI libraries that interfere with qt

Issue #234, opened 2024-04-04
Use cyto-dl overrides manager to generate overrides for prediction


yrkim98:
Training changed to use a CytoOverridesManager class to generate overrides for cyto-dl.
Use similar pattern for Predictions.

PR #625, opened 2025-05-08
Bugfix/enforce single thread osx

yrkim98:
## Purpose
Fixes an issue thao has reported on apple silicon macs- cannot use cyto-dl with `num_workers>0`

## Changes
Need to set `num_workers=0` https://stackoverflow.com/questions/64772335/pytorch-w-parallelnative-cpp206

### DIFF
--- a/src/allencell_ml_segmenter/services/prediction_service.py
+++ b/src/allencell_ml_segmenter/services/prediction_service.py
@@ -1,4 +1,5 @@
 import csv
+from sys import platform
 
 from allencell_ml_segmenter.core.subscriber import Subscriber
 from allencell_ml_segmenter.core.event import Event
@@ -174,6 +175,11 @@ class PredictionService(Subscriber):
             overrides["trainer.accelerator"] = "gpu"
         else:
             overrides["trainer.accelerator"] = "cpu"
+
+        # pytorch needs num_workers = 0 for mac osx
+        # keeps prediction on main thread
+        if platform.system() == "Darwin":
+            overrides["data.num_workers"] = 0
 
         return overrides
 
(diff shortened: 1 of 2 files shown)


src/allencell_ml_segmenter/services/prediction_service.py
```diff
@@ -175,6 +176,11 @@ def build_overrides(self, checkpoint: Path) -> dict[str, Any]:
         else:
             overrides["trainer.accelerator"] = "cpu"
 
+        # pytorch needs num_workers = 0 for mac osx
+        # keeps prediction on main thread
+        if platform.system() == "Darwin":
```
hughes036, line 181: Why do we need this in 2 places?  Does prediction not use `cyto_overrides_manager` ?

hughes036 (approved):

hughes036:
The affect of this is that data will be loaded without any parallelism, but we are talking about MAC where we are already not support GPU and this will just make the unusably bad performance slightly worse, correct?

saeliddp (approved):

src/allencell_ml_segmenter/services/prediction_service.py
yrkim98: yeah prediction does not use `cyto_overrides_manager` and we've had a ticket open to refactor that for a while [here](https://github.com/AllenCell/allencell-segmenter-ml/issues/234)

yrkim98:
> The affect of this is that data will be loaded without any parallelism, but we are talking about MAC where we are already not support GPU and this will just make the unusably bad performance slightly worse, correct?

correct. We need to emphasize this isn't the way we intend users to run our plugin. They could test on a mac but ultimately you need a windows/linux workstation with gpu.

yrkim98:
im running a final test on my computer for this- will merge in a sec and do a release- need like 30 minutes for it to complete.

Issue #627, opened 2025-05-16
Timeout function in Training causing training to fail


thao-do:
**Describe the bug**
In Training tab, if the timeout option box is checked, the training will fail. This is an existing bug since Nov 2024 and reproducible on both Mac & Windows OS.

**To Reproduce**
1. Start a new model
2. Go through Curation & into Training
3. Fill in all training params as normal, AND check the box to "timeout" and fill in the number of minutes to time out
4. Start training - a popup modal would notify user that training has failed

Console logs: 
- [old console log file - Nov 2024](https://docs.google.com/document/d/1k0IquLJHwWwyPbH3PydJb4-3wooYQ7yg_bg1fRGM1Ec/edit?usp=sharing)
- [console log file - Mac](https://docs.google.com/document/d/1ZbHT8NKEzE6zjvURY7di8DfxMW8LWM2lv3Qr1b0v9XI/edit?usp=sharing)
- [console log file - Windows](https://docs.google.com/document/d/1F3iB3b8Zk6NdLwyxoo0jxG97ooP1lkWBmGf0A74j29s/edit?usp=sharing)

**Expected behavior**
Training should run as normal and ends at the timeout specified time.

**Screenshots**
<img width="607" alt="Image" src="https://github.com/user-attachments/assets/0404f024-be9d-4bde-b19e-07c9c083bca9" />

**Describe your data (image format, 2D /3D etc.)** LaminB1 sample dataset

**Environment (please complete the following information):**
 - OS: Mac OS 13.6 (22G120), Windows OS built 20348.2655 (EC2 instance)
 - Plugin Version: 1.0.0rc8
 - PyTorch version 2.0.1 on Windows OS
 - GPU? yes on Windows OS
 - CUDA version [e.g. 10.0]

**Additional context**
We discussed to remove this feature completely, with the conditions that:
1. User should be able to estimate the time base on how long each epoch might potentially take and set the appropriate number of epoch
2. In case training needs to be stopped before it reach the set number of epoch or before the training is auto-stopped when the model performance is no longer improved, user should be able to cancel the training

yrkim98:
Looks like cyto-dl or pytorch may have changed to expect this `max_time` override in a different format. Removed in https://github.com/AllenCell/allencell-segmenter-ml/pull/628, can readd later on if needed. Haven't investigated exactly which format we need since it was removed as a non-priority funcitonality

Issue #469, opened 2024-07-25
Canceling training run while running freezes app


yrkim98:
Investiate + fix canceling a training run through the cancel button on the progress bar- this currently freezes the app.

Originally reported for a test run on Mac, now exists on test run on Windows OS / EC2 instance `winsegml2`

### Step to reproduce: 
1. Start training
2. Once the progress modal shows up, either wait for a bit (let the file cached, or let the first few epochs run) or without waiting, click the [Cancel] button below the progress bar in the modal
3. The progress modal would close, napari & the plugin look like they've returned to normal >> this part makes it look like "cancelling" works
4. Click anywhere on napari, the napari title bar would change to "Not responding"
5. Wait for a long time and the "not responding" is not resolving

thao-do:
Resurface this ticket with some additional details of steps to reproduced on the Windows OS EC2 instance

yrkim98:
I was able to recreate this.

When we cancel a training run, `QThread.terminate()` is called, which does not guarantee the thread can be killed cleanly and if terminated at certain stages of training this causes the app to freeze.

it seems like cancelling training during different steps affects whether the thread terminates successfully or not (for example, if cancelled during setup or caching -> works fine, if cancelled mid epoch -> sometime freezes, sometimes doesnt)

The **best path forward** seems to be to run the training task inside a separate process from within the thread. that process can be killed cleanly without freezing the app. we need to keep the QThread layer (rather than ditching QThread and just using a separate process) since we have signal/slot behavior (on error, on finish) expected by the UI and other parts of the app.

Some stackoverflow posts I looked at:

- https://stackoverflow.com/questions/60939559/qt-thread-execution-does-not-stop-after-exit-quit-or-terminate-was-called
- https://stackoverflow.com/questions/16246796/how-to-stop-a-qthread-from-the-gui
- https://stackoverflow.com/questions/32053618/how-to-to-terminate-process-using-pythons-multiprocessing
- https://stackoverflow.com/questions/7928044/how-to-properly-terminate-a-qthread-from-a-gui-application
- https://www.qtcentre.org/threads/15106-Terminating-QThread-error
- https://stackoverflow.com/questions/49087539/how-to-interrupt-a-script-execution-on-a-qthread-in-python-pyqt

Cancelling was removed in https://github.com/AllenCell/allencell-segmenter-ml/pull/628.

PR #628, opened 2025-05-19
Remove timeout functionality

yrkim98:
## Purpose
Seems like these no longer work- we will remove the timeout and cancel functionality. We can still specify a max number of epochs, which works.

Removing UI code but leaving everything else in case we'd like to fix and add it in the future.

**timeout** stopped working due to a cyto-dl change: looks like we need to specify max time in a different format. [ticket](https://github.com/AllenCell/allencell-segmenter-ml/issues/627)

**canceling a training run** doesnt work because QThread.terminate() cant guarantee the thread stops and doesnt release the python GIL cleanly and causes the entire app to freeze. it seems like cancelling training during different steps affects whether the thread terminates successfully or not (for example, if cancelled during setup or caching -> works fine, if cancelled mid epoch -> sometime freezes, sometimes doesnt-- which is why we previously may have thought it works fine
)
the best path forward seems to be to run the training task inside a separate process from within the thread. that process can be killed cleanly without freezing the app. we need to keep the QThread layer (rather than ditching QThread and just using a separate process) since we have signal/slot behavior (on error, on finish) expected by the UI and other parts of the app. [ticket
](https://github.com/AllenCell/allencell-segmenter-ml/issues/469#issuecomment-2895524115)
## Changes
Remove timeout and cancel functionality from UI

## Testing
Works on windows desktop, tests pass

### DIFF
--- a/src/allencell_ml_segmenter/training/view.py
+++ b/src/allencell_ml_segmenter/training/view.py
@@ -213,36 +213,6 @@ class TrainingView(View, MainWindow):
         )
         bottom_grid_layout.addWidget(self._num_epochs_input, 3, 1)
 
-        max_time_layout: QHBoxLayout = QHBoxLayout()
-        max_time_layout.setSpacing(0)
-
-        self._max_time_checkbox: QCheckBox = QCheckBox()
-        self._max_time_checkbox.setObjectName("timeoutCheckbox")
-        self._max_time_checkbox.stateChanged.connect(
-            self._max_time_checkbox_slot
-        )
-        max_time_layout.addWidget(self._max_time_checkbox)
-
-        max_time_left_text: QLabel = QLabel("Time out after")
-        max_time_layout.addWidget(max_time_left_text)
-
-        self._max_time_in_minutes_input: QLineEdit = QLineEdit()
-        self._max_time_in_minutes_input.setObjectName("timeoutMinuteInput")
-        self._max_time_in_minutes_input.setEnabled(False)
-        self._max_time_in_minutes_input.setMaximumWidth(30)
-        self._max_time_in_minutes_input.setPlaceholderText("30")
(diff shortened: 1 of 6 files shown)


src/allencell_ml_segmenter/training/view.py
```diff
@@ -213,36 +213,6 @@ def __init__(
...
-        )
-        max_time_layout.addWidget(self._max_time_in_minutes_input)
-
-        max_time_right_text: LabelWithHint = LabelWithHint("minutes")
-        max_time_right_text.set_hint("(Optional) Maximum time to train model")
-        max_time_layout.addWidget(
-            max_time_right_text, alignment=Qt.AlignmentFlag.AlignLeft
-        )
-        max_time_layout.addStretch()
-        bottom_grid_layout.addLayout(max_time_layout, 4, 1)
```
saeliddp, line 245: Any theories why timeout stopped working? And is this something we want to revisit in the future? If so, maybe comment these lines out instead of deleting them.

saeliddp (approved):
Curious about the root of these issues, but approving the immediate changes

src/allencell_ml_segmenter/training/view.py
hughes036: @yrkim98 can you document what you told me in Slack about the cause of the breakage?

Please document it somewhere in the repo, as well as in the PR description.

hughes036 (changes requested):
requesting documenting the background - when / how it worked, why we think it broke, what we could do to get it back.

src/allencell_ml_segmenter/training/view.py
yrkim98: pr description updated.

the two existing tickets were updated with the deets.

hughes036 (approved):

PR #635, opened 2025-09-19
Bugfix/fix layers changed deprecation

yrkim98:
## Purpose
This fixes `layer_change` napari event being deprecated. 

Some background- we developed the plugin to respond to a napari `layer_change` event. Napari used to emit this event on any change to the layer list or any items inside napari's layer list, and we used this event to respond to sync our `FileInputWidget` up with the list of images available on the napari layer list. 

Now, napari seems to have gotten rid of the more general `layer_change` event for more fine grained events that fire based on different changes to the layer list.

The four events we need to be aware of are:
- `layers.events.inserted` (image added)
- `layers.events.removed` (image removed)
- `layers.events.moved` (layer reordered)
- `layers.events.changed` (an existing layer modified)

## Changes
I made the smallest change possible, which was directly replacing the `layer_change` event with the four events above that the plugin needs to be aware of when responding to changes in the layer list. 

A more ideal change would be to create dedicated handlers which respond to these changes individually, but this is a much bigger refactor. for now this keeps the plugin happy with the latest version of napari

Since we made this change, we can no longer use `layer_change` events and we have to drop support for all napari versions `< 0.6.2`, which means we need to drop python 3.9 support as well (since `napari==0.6.2` only supports python 3.10 and above)

## Testing
Tested by installing fresh on windows, and running a full curation -> training -> prediction -> thresholding run.

## How to review
Small change

### DIFF
--- a/.github/workflows/test_lint.yml
+++ b/.github/workflows/test_lint.yml
@@ -43,6 +43,12 @@ jobs:
       - name: Type check with mypy
         run: |
           mypy src/allencell_ml_segmenter
+      # This step ensures that opencv does not install libraries
+      # that may conflict with what is needed to run pyqt unit tests
+      - name: Force opencv-python-headless
+        run: |
+          pip uninstall -y opencv-python
+          pip install "numpy<2.0" opencv-python-headless
       - name: Test with pytest
         run: |
           pytest -v --color=yes --cov=allencell_ml_segmenter --cov-report=xml
--- a/.github/workflows/test_lint.yml
+++ b/.github/workflows/test_lint.yml
@@ -48,7 +48,7 @@ jobs:
       - name: Force opencv-python-headless
         run: |
           pip uninstall -y opencv-python
-          pip install opencv-python-headless
+          pip install "numpy<2.0" opencv-python-headless
       - name: Test with pytest
         run: |
           pytest -v --color=yes --cov=allencell_ml_segmenter --cov-report=xml
--- a/.github/workflows/test_lint.yml
+++ b/.github/workflows/test_lint.yml
@@ -43,12 +43,6 @@ jobs:
       - name: Type check with mypy
         run: |
           mypy src/allencell_ml_segmenter
-      # This step ensures that opencv does not install libraries
-      # that may conflict with what is needed to run pyqt unit tests
-      - name: Force opencv-python-headless
-        run: |
-          pip uninstall -y opencv-python
-          pip install "numpy<2.0" opencv-python-headless
       - name: Test with pytest
         run: |
           pytest -v --color=yes --cov=allencell_ml_segmenter --cov-report=xml
--- a/docs/user_docs/conf.py
+++ b/docs/user_docs/conf.py
@@ -10,7 +10,7 @@
 project = "Segmenter ML Plugin for napari"
 copyright = "2024, Allen Institute for Cell Science"
 author = "Segmenter ML plugin team"
-release = "1.0.1"
+release = "1.0.1rc1"
 
 
 # -- General configuration ---------------------------------------------------
--- a/docs/user_docs/conf.py
+++ b/docs/user_docs/conf.py
@@ -10,7 +10,7 @@
 project = "Segmenter ML Plugin for napari"
 copyright = "2024, Allen Institute for Cell Science"
 author = "Segmenter ML plugin team"
-release = "1.0.1dev0"
+release = "1.0.1rc0"
 
 
 # -- General configuration ---------------------------------------------------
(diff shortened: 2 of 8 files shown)


yrkim98:
test pass locally but fail on ci, checking

src/allencell_ml_segmenter/main/viewer.py
```diff
@@ -121,7 +121,12 @@ def get_layers_nonthreshold(self) -> list[Layer]:
     def subscribe_layers_change_event(
         self, function: Callable[[NapariEvent], None]
     ) -> None:
-        self.viewer.events.layers_change.connect(function)
```
saeliddp, line 124: is `layers_change` still a valid event (just unused) in the newer napari versions? If so, I would leave this in for backwards compatibility with older napari versions. If not, we may want to introduce a minimum napari version in our requirements--whatever version started using these new events instead

yrkim98: it is no longer valid at all. Unfortunately napari needs to be installed separately from our deps (napari is not in our pyproject deps list) so we can only enforce it through instructions on our user guide- which i will update now.

(I wonder if there's a better way to do this though, napari apps that appear on napari hub don't list napari as a dependancy, but since ours is not on napari hub and will not be in the future maybe we can add that back in)

yrkim98: I pinned `napari >= 0.6.2` since we have this change, and also dropped support for python 3.9 since 0.6.2 supports 3.10+ only

saeliddp: > but since ours is not on napari hub and will not be in the future maybe we can add that back in

I don't have context for why this is the case, but seems like this means pinning should be fine.

Another (kinda messy) solution would be to just try/except connecting to the different events.

hughes036: Is there a superclass of both `layer_change` and the new events that could be handled?

docs/user_docs/conf.py
```diff
@@ -10,7 +10,7 @@
 project = "Segmenter ML Plugin for napari"
 copyright = "2024, Allen Institute for Cell Science"
 author = "Segmenter ML plugin team"
-release = "1.0.0"
+release = "1.0.1rc0"
```
hughes036: was this change made by bumpver?  dont we have a pattern of merging this separate from the bugfix?

.github/workflows/test_lint.yml
```diff
@@ -48,7 +48,7 @@ jobs:
       - name: Force opencv-python-headless
         run: |
           pip uninstall -y opencv-python
-          pip install opencv-python-headless
+          pip install "numpy<2.0" opencv-python-headless
```
hughes036, line 51: how does this relate to the pr purpose?

src/allencell_ml_segmenter/_tests/main/test_main_widget.py
```diff
@@ -155,10 +171,14 @@ def test_experiments_home_initialized(qtbot: QtBot) -> None:
     settings.set_user_experiments_path(
         None
     )  # Simulates state where users has not yet chosen an experiments home.
+    viewer = Mock(
+        spec="napari.Viewer"
+    )  # set up fake viewer with layers that can emit events
+    viewer.layers = FakeLayers()
```
hughes036, line 177: not for stubbing?  does the fake participate in asserts?

hughes036 (approved):
see comments about event super and using new fake more effectively.

src/allencell_ml_segmenter/main/viewer.py
yrkim98: @hughes036 no it was just removed completely https://github.com/napari/napari/pull/7971/files

docs/user_docs/conf.py
yrkim98: we havent used bumpver for this project in a while, and I did this to do a fresh install + test on windows

.github/workflows/test_lint.yml
yrkim98: > Since we made this change, we can no longer use layer_change events and we have to drop support for all napari versions < 0.6.2, which means we need to drop python 3.9 support as well (since napari==0.6.2 only supports python 3.10 and above)

dropping python 3.9 meant that pip automatically now grabs numpy 2.x.x -> we need this pin to keep things working

src/allencell_ml_segmenter/_tests/main/test_main_widget.py
yrkim98: it doesnt participate in any asserts, just keeps `MainWidget` happy on tests

saeliddp (approved):

